PR #1, opened 2018-11-19
Print the current logged in user

mikethebeer:
This adds the functionality to print the user that is currently logged in on the CrateDB cloud platform.
For now the default output format of the result is `json`, e.g.

```json
$ croud me -r bregenz.a1 --env dev -of json
{
  "email": "michael@crate.io",
  "username": "Google_104209952803880732135",
  "name": "Michael Beer"
}
```

Since the built-in OAuth2 mechanism using `croud login` is not in place yet I introduced the env var `CLOUD_SESSION` which can be used to test and simulate the login on CrateDB Cloud (check README for detailed usage). This env **could** be removed in the recent future when we have `croud login` in place.


setup.py
```diff
@@ -37,16 +37,17 @@
...
-    install_requires=["argh"],
+    install_requires=["argh", "requests", "colorama"],
     extras_require={
         "testing": [
             "pytest>=3",
-            "pytest-flake8==1.0.0.",
-            "pytest-isort==0.2.1",
-            "black==18.6b2",
-            "pytest-black==0.1.6",
+            "pytest-flake8",
```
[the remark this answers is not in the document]
MarkusH: I usually only pin version for test dependencies when there are changes that I can't "just fix" and need more time for. So, for now, I'd go with an unpinned version. And if need be add a `<x.y.z`.

mikethebeer:
retest this please

mikethebeer:
retest this please

croud/me.py
```diff
@@ -17,10 +17,36 @@
...
+
+@argh.arg(
+    "-r",
+    "--region",
+    choices=["westeurope.azure", "eastus.azure", "bregenz.a1"],
+    default="bregenz.a1",
+    type=str,
+)
+@argh.arg("--env", choices=["dev", "prod"], default="prod", type=str)
+@argh.arg("-of", "--output-fmt", choices=["json"], default="json", type=str)
```
MarkusH: Having a short-command flag (starting with a single `-`) with more than one letter is uncommon. POSIX recommends only using a single, alpha-numeric character: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

croud/gql.py
```diff
@@ -0,0 +1,53 @@
...
+    if env.lower() == "dev":
+        host = CLOUD_DEV_HOST
+
+    request = requests.post(
+        f"https://{region}.{host}/graphql",
+        json={"query": query},
+        cookies=dict(session=session),
+    )
+
+    if request.status_code == 200:
```
MarkusH: Do you want to check for a `200` or for "a successful request"? In case of the latter, maybe consider using `if request.ok:`

tests/unit_tests/test_commands.py
```diff
@@ -0,0 +1,39 @@
...
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
+# License for the specific language governing permissions and limitations
+# under the License.
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+import unittest
+from unittest.mock import patch
```
MarkusH: This one might be worth considering: https://stackoverflow.com/questions/49271586/valueerror-wrapper-loop-when-unwrapping . See also: https://github.com/crate/cloud-app/pull/478#discussion_r233495977

PR #7, opened 2018-11-23
Added logout and removed --env flag requirements

alijared:


mikethebeer:
@alijared pls run `black .` to get rid of the linting errors.

MarkusH:
Doesn't this change essentially remove the ability to be logged-in to dev and prod simultaneously?

mikethebeer:
@alijared btw I get this when I want to use it

```
(env) ☁  croud [aa/logout] croud -h                        
Traceback (most recent call last):
  File "/c/sandbox/crate/croud/env/bin/croud", line 11, in <module>
    load_entry_point('croud', 'console_scripts', 'croud')()
  File "/c/sandbox/crate/croud/croud/__main__.py", line 34, in main
    Configuration.create()
  File "/c/sandbox/crate/croud/croud/config.py", line 34, in create
    write_config({'env': 'prod', 'token': ''})
  File "/c/sandbox/crate/croud/croud/config.py", line 58, in write_config
    with open(Configuration.FILEPATH, 'w', encoding='utf8') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/mibe/.config/crate/croud/config.yaml'
```

croud/config.py
```diff
@@ -18,27 +18,52 @@
...
 
 import os
-from pathlib import Path
 
+import yaml
 from appdirs import user_config_dir
 
 
 class Configuration:
+    USER_CONFIG_DIR: str = user_config_dir("crate/croud")
```
mikethebeer: CraSh stores its configuration inside `.config/Crate/crash.cfg` 👉 https://github.com/crate/crash/blob/10e98fec348ede7029268f6c201ba029e06d05eb/src/crate/crash/command.py#L69

I think it would be better to store it in the same folder

croud/session.py
```diff
@@ -69,6 +68,14 @@ def __init__(
                 f"Query failed to run by returning code of {resp.status}. {query}"
             )
 
+    async def logout(self):
+        domain = "cratedb.cloud"
```
mikethebeer: pls use the already defined constants `CLOUD_PROD_DOMAIN` and `CLOUD_DEV_DOMAIN` instead

croud/config.py
alijared: Changed

croud/session.py
alijared: Changed

PR #8, opened 2018-11-26
Add tabular output format

mikethebeer:
```
$ croud me --env dev -o tabular
+-----------------+------------------------------+-----------------+
| email           | username                     | name            |
|-----------------+------------------------------+-----------------|
| hoschi@crate.io | Google_104209952803880732135 | Hoschi Galoschi |
+-----------------+------------------------------+-----------------+
```

### DIFF
--- a/croud/printer.py
+++ b/croud/printer.py
@@ -18,15 +18,17 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import json
+from typing import List, Union
 
 from colorama import Fore, Style
+from tabulate import tabulate
 
 from croud.typing import JsonDict
 
 
-def print_format(res: JsonDict, format: str = "json") -> None:
+def print_format(rows: Union[List[JsonDict], JsonDict], format: str = "json") -> None:
     printer = FormatPrinter()
-    printer.print_resultset(res, format)
+    printer.print_rows(rows, format)
 
 
 def print_error(text: str):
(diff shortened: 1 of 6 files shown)


tests/unit_tests/test_printer.py
```diff
@@ -17,17 +17,47 @@
...
+|-----+-----+------|
+| foo |   1 | TRUE |
++-----+-----+------+
+"""
+        )
+
+    def test_tabular_format_multi_row(self, capsys):
+        sample = [
+            {"a": "foo", "b": 1, "c": True},
+            {"a": {"bar": [{"value": 0}, {"value": 1}]}, "b": 2, "c": False},
```
MarkusH: Can you change the order of keys in this line. Python 3.7 ensures insertion order in dictionaries and mixing up the order here would ensure that the mapping of column to keys works correctly:
```suggestion
            {"b": 2, "c": False, "a": {"bar": [{"value": 0}, {"value": 1}]}},
```

croud/printer.py
```diff
@@ -39,15 +41,48 @@ def print_info(text: str):
...
+            # +-----+-----+
+            # | bar |   2 |
+            # +-----+-----+
+            headers = list(map(str, iter(rows[0].keys())))
+            values = [
+                [self._transform_field(row[header]) for header in headers]
+                for row in rows
+            ]
+        else:
+            values = [list(map(self._transform_field, rows.values()))]
```
MarkusH: I'd move the `values` assignment after the `headers` assignment to have the same order as above (lines 80/81). Other than that :+1:

mikethebeer: @MarkusH okay. pushed a fixup 👇

PR #10, opened 2018-11-26
Added env command and refactored config to use multi-auth

alijared:


croud/util.py
```diff
@@ -76,3 +76,14 @@ def open_page_in_browser(url: str) -> int:
...
     return webbrowser.open_new_tab(url)
+
+
+def nested_dict_value(data: dict, key: str):
+    cursor = data
+    for k in key.split("."):
+        if k in cursor:
+            cursor = cursor[k]
+        else:
+            return None
```
MarkusH: I feel like this should raise a `KeyError` or something instead of only returning `None`.

croud/config.py
```diff
@@ -22,35 +22,59 @@
...
 
+from croud.util import nested_dict_value
+
 
 class Configuration:
     USER_CONFIG_DIR: str = user_config_dir("Crate")
     FILENAME: str = "croud.yaml"
     FILEPATH: str = f"{USER_CONFIG_DIR}/{FILENAME}"
+    DEFAULT_CONFIG: dict = {
+        "auth": {"currentContext": 0, "contexts": [{"env": "prod", "token": ""}]}
```
MarkusH: Could you use snake_case over camelCase please :)
```suggestion
        "auth": {"current_context": 0, "contexts": [{"env": "prod", "token": ""}]}
```

tests/unit_tests/test_util.py
```diff
@@ -86,3 +87,8 @@ def test_can_launch_browser(self, webbrowser_get_mock, get_platform_mock):
             env_mock.get.return_value = "foo"
             result = can_launch_browser()
             self.assertFalse(result)
+
+    def test_nested_dict_value(self):
```
MarkusH: Can you add a test for the absence of a nested key as well as when the value of a key isn't a dictionary (or rather, doesn't support key lookup).

Something  like this:
```python
data = {"key": {"nestedkey": {"reallynestedkey": "thenestedvalue"}}}
key = nested_dict_value(data, "key.nestedkey.doesnotexist")
```
as well as
```python
data = {"key": {"nestedkey": 1}}
key = nested_dict_value(data, "key.nestedkey.foo")
```

tests/unit_tests/test_util.py
```diff
@@ -86,3 +87,8 @@ def test_can_launch_browser(self, webbrowser_get_mock, get_platform_mock):
...
+        data = {"key": {"nestedkey": {"reallynestedkey": "thenestedkey"}}}
```
MarkusH:
```suggestion
        data = {"key": {"nestedkey": {"reallynestedkey": "thenestedvalue"}}}
```

croud/config.py
```diff
@@ -63,7 +87,7 @@ def write_config(config: dict) -> None:
         yaml.dump(config, f, default_flow_style=False, allow_unicode=True)
 
 
-def set_property(property: str, value: str):
```
MarkusH: I believe there's still merit to keep this method. There will be other commands that make use of this, I think. At least the `login` command, I think.

CHANGES.rst
```diff
@@ -14,3 +14,5 @@ Unreleased
 - Added a subcommand `logout` that clears the stored login token, and clears the OAuth session
 
 - The environment used to logged to in is now stored in config, so it is only necessary to provide the ``--env`` flag on the ``login`` subcommand
+
+- Added a subcommand `env` that allows you to switch env (so you can be logged into multiple environments. E.g. ``croud env prod``
```
jonas-arkulpa: We normally add new changes at the top of the list.

croud/config.py
```diff
@@ -22,35 +22,59 @@
...
             os.makedirs(Configuration.USER_CONFIG_DIR)
 
         if not os.path.isfile(Configuration.FILEPATH):
-            write_config({"env": "prod", "token": ""})
+            write_config(Configuration.DEFAULT_CONFIG)
 
     @staticmethod
     def get_env() -> str:
-        return load_config().get("env") or "prod"
+        return get_auth_context().get("env") or "prod"
```
MarkusH: Why don't we do this?
```suggestion
        return get_auth_context().get("env", "prod")
```

jonas-arkulpa: Would it make sense to also allow this config via env variable?
Do we have this functionality for the region?

croud/config.py
```diff
@@ -22,35 +22,59 @@
...
 
     @staticmethod
-    def set_env(env: str) -> None:
-        set_property("env", env)
+    def set_context(env: str) -> None:
+        config = load_config()
+        contexts: list = nested_dict_value(config, "auth.contexts")
```
MarkusH: I think `auth["contexts"]` should be a dict like structure and `current_context` point to the key of the current context:
```python
{
    "auth": {
        "current_context": "dev",  # or "prod"
        "contexts": {
            "dev": {"token": ...},
            "prod": {"token": ...},
        },
    },
}
```

croud/config.py
alijared: What do you mean? This returns the env for the current context. To change you can ``croud env $env``

tests/unit_tests/test_util.py
```diff
@@ -86,3 +87,20 @@ def test_can_launch_browser(self, webbrowser_get_mock, get_platform_mock):
...
+        key = nested_dict_value(data, "key.nestedkey.reallynestedkey")
+        self.assertEqual(key, "thenestedkey")
+
+        data = {"key": {"nestedkey": {"reallynestedkey": "thenestedvalue"}}}
+        with self.assertRaises(KeyError) as error:
+            nested_dict_value(data, "key.nestedkey.doesnotexist")
+
+        self.assertIsInstance(error.exception, KeyError)
```
MarkusH: No need for `assertIsInstance` checks on `error.exception`. That's already handled by the `self.assertRaises()` :slightly_smiling_face:

croud/util.py
```diff
@@ -76,3 +76,16 @@ def open_page_in_browser(url: str) -> int:
...
             pass
     return webbrowser.open_new_tab(url)
+
+
+def nested_dict_value(data: dict, key: str):
+    cursor = data
+    for k in key.split("."):
+        if type(cursor) is not dict:
+            raise TypeError
+        if k in cursor:
```
MarkusH: No need to do a check here. Just using `cursor = cursor[key]` will raise the `KeyError` for if `key` is not in `cursor`. Unless you want to provide additional information on the `KeyError`, which might be helpful.

croud/util.py
```diff
@@ -76,3 +76,16 @@ def open_page_in_browser(url: str) -> int:
         except FileNotFoundError:  # WSL might be too old
             pass
     return webbrowser.open_new_tab(url)
+
+
+def nested_dict_value(data: dict, key: str):
+    cursor = data
+    for k in key.split("."):
+        if type(cursor) is not dict:
+            raise TypeError
```
MarkusH: I think it would be valuable to provide a detailed error message what type `cursor` actually is

croud/util.py
```diff
@@ -76,3 +76,16 @@ def open_page_in_browser(url: str) -> int:
         except FileNotFoundError:  # WSL might be too old
             pass
     return webbrowser.open_new_tab(url)
+
+
+def nested_dict_value(data: dict, key: str):
+    cursor = data
+    for k in key.split("."):
+        if type(cursor) is not dict:
```
MarkusH: The more pythonic way would be `if not isinstance(cursor, dict)`

tests/unit_tests/test_util.py
alijared: Thanks

jonas-arkulpa:
I think it would be enough to save the credentials for the env separately and spare the sub command.
But we discussed the additional value of this feature the way it is implemented now.

@mikethebeer It's currently already an optional argument, right?

If it helps developers I don't mind the extra command.

mikethebeer:
> @mikethebeer It's currently already an optional argument, right?

On the `me` subcommand, this got reomoved with https://github.com/crate/croud/pull/7 . However, we use it already in `croud login --env dev` as optional argument. So that is why I tend to do it for the other commands as well like this.

alijared:
I disagree about the subcommand for ``env`` because as you said, it's for developers only. I don't want to have to specify an ``--env`` flag every time, and write an optional argument for every command we write. As a developer, I want to switch environments, not parse it on every command - that's cumbersome and unnecessary. If there's a way to add the optional argument to every command without having to repeat yourself, then that's not so bad, but I'd still have to either login again to work in a different environment, or parse ``--env`` on every command - which is fine now as we have one command, but in the future we won't. 

In any case, the env subcommand was 3 lines of code, and can be removed in favour of the always optional flag

croud/env.py
```diff
@@ -0,0 +1,26 @@
...
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+from croud.config import Configuration
+from croud.printer import print_info
+
+
+def env(env: str):
+    Configuration.set_context(env)
+    print_info(f'Environment switched to "{env}"')
```
mikethebeer:
```suggestion
    print_info(f"Environment switched to {env}")
```

CHANGES.rst
```diff
@@ -5,6 +5,12 @@ Changes for croud
 Unreleased
 ==========
 
+- Added a tabular format to the list of possible output formats in subcommands.
```
mikethebeer: thats not a change u made in this commit. Probably a rebase issue

CONTRIBUTING.rst
```diff
@@ -0,0 +1,10 @@
+============
```
mikethebeer: please rebase

croud/env.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# However, if you have executed another commercial license agreement
+# with Crate these terms will supersede the license and you may use the
+# software solely pursuant to the terms of the relevant commercial agreement.
+
+from croud.config import Configuration
+from croud.printer import print_info
+
+
+def env(env: str):
```
mikethebeer: Pydoc missing --> so I think `croud env -h` prints nothing right now

croud/env.py
```diff
@@ -0,0 +1,26 @@
...
+    Configuration.set_context(env)
```
mikethebeer: what happens if a user sets an irregular env? I think you need to restrict to `dev` or `prod` somewhere. Probably already in the parser?

croud/me.py
```diff
@@ -34,7 +35,8 @@
     type=str,
 )
 @argh.arg("-o", "--output-fmt", choices=["json"], default="json", type=str)
-def me(region=None, output_fmt=None) -> None:
+@argh.arg("--env", default=None, type=str)
```
mikethebeer: If you go with the optional argument `env` it would make sense to let the user chose between `dev` and `prod` to avoid irregular envs

croud/env.py
mikethebeer: Still missing

croud/env.py
mikethebeer: ☝️

croud/session.py
```diff
@@ -58,7 +58,7 @@ def __init__(
             cookies={"session": token}, connector=conn, headers=headers
         )
 
-    async def fetch(self, query: str) -> Union[List[JsonDict], JsonDict]:
+    async def fetch(self, query: str) -> JsonDict:
```
mikethebeer: why this change? I think its correct the way it is since graphql could return also a list of dictionionaries

croud/util.py
```diff
@@ -76,3 +76,11 @@ def open_page_in_browser(url: str) -> int:
         except FileNotFoundError:  # WSL might be too old
             pass
     return webbrowser.open_new_tab(url)
+
+
+def nested_dict_value(data: dict, key: str):
```
mikethebeer: as @MarkusH already point out, why is this method necessary? why don't you access the values of a dict using subscript syntax e.g. `mydict["a"]["b"]`.

croud/config.py
```diff
@@ -22,48 +22,88 @@
...
-            write_config({"env": "prod", "token": ""})
+            write_config(Configuration.DEFAULT_CONFIG)
 
     @staticmethod
     def get_env() -> str:
-        return load_config().get("env") or "prod"
+        if Configuration._using_context is not None:
+            return Configuration._using_context
+
+        return nested_dict_value(load_config(), "auth.current_context")
```
mikethebeer: as questioned already. Is `nested_dict_value` needed just for readability. Why don't you access it with `cfg["auth"]["current_context"]`

croud/session.py
alijared: Fails tests with ``Union[List[JsonDict], JsonDict]`` return type

croud/config.py
alijared: Because this is cleaner, and could be used in the future for other dicts that have nested keys

croud/session.py
mikethebeer: but then I think the code that calls `fetch` need to adapt its types.

croud/config.py
chaudum: In you really have "problems" with nested data structures, you should think about using for example `addict` https://github.com/mewwts/addict

croud/config.py
```diff
@@ -70,17 +67,17 @@ def get_token() -> str:
     def set_token(token: str) -> None:
         config = load_config()
 
-        context = Configuration._using_context
-        if context is None:
-            context = nested_dict_value(config, "auth.current_context")
+        context = Configuration.current_context
+        if context == "":
```
mikethebeer: use

```python
if not context:
    ...
```

croud/config.py
```diff
@@ -102,8 +99,8 @@ def write_config(config: dict) -> None:
 def get_auth_context() -> dict:
     config = load_config()
 
-    context = Configuration._using_context
-    if context is None:
-        context = nested_dict_value(config, "auth.current_context")
+    context = Configuration.current_context
+    if context == "":
```
mikethebeer: same here:

```python
if not context:
```

CHANGES.rst
```diff
@@ -5,6 +5,10 @@ Changes for croud
 Unreleased
 ==========
 
+- Added ``env`` flag to commands to temporarily override auth context
+
+- Added a subcommand `env` that allows you to switch env (so you can be logged into multiple environments. E.g. ``croud env prod``
```
chaudum: double backtick around `env`

PR #11, opened 2018-11-26
[PROPOSAL] Load commands from `croud_commands` entry points.

MarkusH:
This allows one to extend the functionality of croud by providing custom commands.


MarkusH:
TIL, `EntryPoint` has a `load()` method :wink:

PR #15, opened 2018-11-28
update installation instructions in readme

chaudum:


README.rst
```diff
@@ -4,19 +4,26 @@ Croud
...
-=======
+Installation
+============
+
+Using pip
+---------
+
+``croud`` can be installed using pip_::
+
+    $ pip install -U croud
```
mfussenegger: I think we might want to start advertising using `python -m pip install` - Using `pip` can lead to the wrong python version being used. Maybe also have the `--local` flag in there as system wide installation is usually bad

chaudum: I'm uncertain about the level of depth of the installation instruction. Of course there are tons of tools that help managing packages and deal with version/dependency problems.
We also tell people in the docs that CrateDB can be installed using plain Linux package manager, but in reality you use provisioning tools (Salt,...) in production.

mfussenegger: I agree that it shouldn't be too detailed and I wouldn't refer to something like pipsi.
Users familiar with python packaging will know how they can make use of packages available via pypi.

For others I think a `python3 -m pip install --user -U croud` is the least harmful. (It avoids them messing up their system installations, and they don't accidentally use a python2 pip or something)

chaudum:
@mfussenegger addressed your concerns in a fixup

PR #19, opened 2018-12-03
Removed name field from gql query in me sub command

alijared:
It was decided that for the time being the CLI should not return the name field of a user, as it may be empty; so I've removed it from the GraphQL query

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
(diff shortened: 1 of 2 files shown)


CHANGES.rst
```diff
@@ -8,6 +8,8 @@ Unreleased
 0.1.0 - 2018/11/28
 ==================
 
+- Removed name field from `me` subcommand
```
mikethebeer: move it to `unreleased`

PR #20, opened 2018-12-03
Added namespaces template

alijared:
In order to have subcommands for entities it's necessary to set up namespaces for argh

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -23,14 +23,70 @@
 
 import colorama
 
+from croud.cmd import CMD, output_fmt_arg, region_arg
 from croud.config import Configuration
-from croud.cmd import CMD
+from croud.env import env
+from croud.login import login
+from croud.logout import logout
+from croud.me import me
+from croud.projects.list import projects_list
+
 
 def main():
     Configuration.create()
     colorama.init()
 
-    CMD()
+    commands: dict = {
+        "me": {
+            "description": "Prints the current logged in user",
+            "usage": "croud me [-h] [--env {prod,dev}]\n\t\t"
+            "[-r {westeurope.azure,eastus.azure,bregenz.a1}] "
+            "[-o {json}]",
+            "extra_args": [region_arg, output_fmt_arg],
+            "calls": me,
+        },
+        "login": {
+            "description": "Performs an OAuth2 Login to CrateDB Cloud",
+            "usage": "croud login [-h] [--env {prod,dev}]",
+            "calls": login,
+        },
+        "logout": {
+            "description": "Performs a logout of the current logged in user",
+            "usage": "croud logout [-h] [--env {prod,dev}]",
+            "calls": logout,
+        },
+        "env": {
+            "description": "Switches auth context",
+            "usage": "croud env [-h] {prod,dev}",
+            "sub_commands": {
+                "prod": {
+                    "description": "Switch auth context to prod",
+                    "usage": "croud env prod",
+                    "calls": env("prod"),
+                },
+                "dev": {
+                    "description": "Switch auth context to dev",
+                    "usage": "croud env dev",
+                    "calls": env("dev"),
+                },
+            },
+        },
+        "projects": {
+            "description": "Project sub commands",
+            "usage": "croud projects [-h] {list,create}",
+            "sub_commands": {
+                "list": {
+                    "description": "Lists all projects for the current "
+                    "user in the specified region",
+                    "usage": "croud projects list [-h] [--env {prod,dev}] [-o {json}]",
+                    "extra_args": [output_fmt_arg],
+                    "calls": projects_list,
+                }
+            },
+        },
+    }
+    CMD(commands)
+
 
 if __name__ == "__main__":
     main()
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -62,12 +62,12 @@ def main():
                 "prod": {
                     "description": "Switch auth context to prod",
                     "usage": "croud env prod",
-                    "calls": env,
+                    "calls": env("prod"),
                 },
                 "dev": {
                     "description": "Switch auth context to dev",
                     "usage": "croud env dev",
-                    "calls": env,
+                    "calls": env("dev"),
                 },
             },
         },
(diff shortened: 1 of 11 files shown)


croud/__main__.py
```diff
@@ -23,10 +23,13 @@
...
 import argh
 import colorama
-import pkg_resources
 
 from croud import __version__
 from croud.config import Configuration
+from croud.login import login
+from croud.logout import logout
+from croud.me import me
+from croud.projects.list import list
```
mikethebeer: would name it to `projects_list` to avoid possible conflict with a keyword

croud/me.py
```diff
@@ -35,7 +35,7 @@
     type=str,
 )
 @argh.arg("-o", "--output-fmt", choices=["json"], default="json", type=str)
-@argh.arg("--env", choices=["prod", "env"], default=None, type=str)
+@argh.arg("--env", choices=["prod", "dev"], default=None, type=str)
```
mikethebeer: this needs to be fixed in a separate hotfix PR

croud/config.py
```diff
@@ -49,7 +49,7 @@ def create() -> None:
 
     @staticmethod
     def get_env() -> str:
-        if not Configuration.current_context:
+        if Configuration.current_context:
```
mikethebeer: hotfix pr

croud/projects/list.py
```diff
@@ -0,0 +1,37 @@
...
+    type=str,
+)
+@argh.arg("-o", "--output-fmt", choices=["json"], default="json", type=str)
+@argh.arg("--env", choices=["prod", "dev"], default=None, type=str)
+def list(region=None, output_fmt=None, env=None) -> None:
+    """
+    Lists all projects for the current user in the specified region
+    """
+
+    # Todo: Return list of projects for logged in user in specified region
```
mikethebeer: To be more pythonic 👉 `pass`
```suggestion
    # Todo: Return list of projects for logged in user in specified region
    pass
```

mfussenegger:
Btw. maybe we should consider moving away from `argh`. Based on my experience using it for cr8 I probably wouldn't use it anymore as it's difficult to compose argument definitions.

We're already starting to run into this problem here that we've multiple sub-commands with the same options. For example `@argh.arg("--env", choices=["prod", "dev"], default=None, type=str)`

Going with plain argparse would make it possible to extract the definitions into re-usable functions. E.g. `add_env_option(parser)`. That way we would ensure that the choices, description and defaults are consistent across all commands.

mikethebeer:
@mfussenegger yes, I had already a bad feeling to start with `argh` here. I would say we could already start switching to the builtin `argparse`. @alijared would it be possible that you make the switch to `argparse` here (in this PR) already?

alijared:
@mfussenegger I agree this approach would be better. @mikethebeer I will update this PR

croud/__main__.py
```diff
@@ -21,28 +21,71 @@
...
-    commands = [
-        entry_point.load()
-        for entry_point in pkg_resources.iter_entry_points("croud_commands")
-    ]
-    p.add_commands(commands)
-    p.dispatch()
+    commands: dict = {
+        "me": {
+            "description": "Prints the current logged in user",
+            "usage": "croud me [-h] [--env {prod,dev}]\n\t\t"
```
mikethebeer: why `\n\t\t` ?

croud/__main__.py
```diff
@@ -21,28 +21,71 @@
...
+            "[-r {westeurope.azure,eastus.azure,bregenz.a1}] "
+            "[-o {json}]",
+            "extra_args": [region_arg, output_fmt_arg],
+            "calls": me,
+        },
+        "login": {
+            "description": "Performs an OAuth2 Login to CrateDB Cloud",
+            "usage": "croud login [-h] [--env {prod,dev}]",
```
mikethebeer: is it possible to extract the usage description of the `default_args` like `--env` to make it reusable?

croud/cmd.py
```diff
@@ -0,0 +1,89 @@
...
+from argparse import ArgumentParser, Namespace
+from typing import Callable
+
+
+class CMD:
+    def __init__(self, commands: dict):
+        parser: ArgumentParser = argparse.ArgumentParser()
+        self.create_parent_cmd(parser, 1, commands)
+
+    def create_parent_cmd(self, parser: ArgumentParser, depth: int, commands: dict):
```
mikethebeer: missing tests

croud/__main__.py
alijared: Formatting. Otherwise the line is very, very long. It's what argparse does for you if you don't specify a custom `usage` field

croud/__main__.py
alijared: It is, but for some reason argparse doesn't provide the accurate usage, instead it only provides accurate positional and optional arguments. So for login, as an example, the standard usage is `croud [-h] [--env {prod,dev}]`. This doesn't show that the `login` argument is supposed to be specified, which doesn't make sense to me, so I overrode them all

alijared: Managed to figure this out. You don't need to specify the usage anymore

setup.py
```diff
@@ -35,15 +35,7 @@
...
-            "env = croud.env:env",
-            "login = croud.login:login",
-            "logout = croud.logout:logout",
-            "me = croud.me:me",
-        ],
-    },
+    entry_points={"console_scripts": ["croud = croud.__main__:main"]},
     packages=["croud"],
     install_requires=[
         "argh",
```
mikethebeer: can be removed

tests/unit_tests/test_commands.py
```diff
@@ -75,20 +75,7 @@ class TestLogout(unittest.TestCase):
...
         m = mock.mock_open()
         with mock.patch("croud.config.open", m, create=True):
-            logout()
+            logout(Namespace(env="dev"))
 
         mock_set_token.assert_called_once_with("")
         mock_print_info.assert_called_once_with("You have been logged out.")
-
-
-class TestEnv(unittest.TestCase):
```
mikethebeer: why did you remove this test?

mikethebeer: ok understand, thx

PR #21, opened 2018-12-05
Note the incompatible storage format of config across versions

mikethebeer:
If the config file that holds the login credentials has an
incompatible or incorrect format the user will be informed to remove the
config file.

```
==> Error: Incompatible storage format in /home/mibe/.config/Crate/croud.yaml. Please remove the config and try again.
```

### DIFF
--- a/croud/config.py
+++ b/croud/config.py
@@ -21,6 +21,13 @@ import os
 
 import yaml
 from appdirs import user_config_dir
+from schema import Schema, SchemaError
+
+from croud.printer import print_error
+
+
+class IncompatibleConfigException(Exception):
+    pass
 
 
 class Configuration:
@@ -35,6 +42,24 @@ class Configuration:
     }
 
     current_context: str = ""
+
+    @staticmethod
+    def validate(config: dict) -> dict:
+        schema = Schema(
+            {
+                "auth": {
+                    "current_context": str,
+                    "contexts": {"prod": {"token": str}, "dev": {"token": str}},
+                }
+            }
+        )
+        try:
+            return schema.validate(config)
+        except SchemaError as e:
+            raise IncompatibleConfigException(
+                f"Incompatible storage format in {Configuration.FILEPATH}. "
+                "Please remove the config and try again."
+            ) from e
 
     @staticmethod
     def create() -> None:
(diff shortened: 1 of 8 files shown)


croud/config.py
```diff
@@ -84,8 +88,14 @@ def override_context(env: str) -> None:
 
 
 def load_config() -> dict:
-    with open(Configuration.FILEPATH, "r") as f:
-        return yaml.load(f) or {}
+    try:
+        with open(Configuration.FILEPATH, "r") as f:
```
chaudum: Reading and parsing the yaml file should never fail. If it fails, it's not because of the incompatibility.

mikethebeer: 🤦‍♂️  OMG, stupid mistake 👍

mikethebeer:
@chaudum repushed the whole work

croud/config.py
```diff
@@ -36,6 +43,24 @@ class Configuration:
...
+                }
+            }
+        )
+        try:
+            return schema.validate(config)
+        except SchemaError as e:
+            raise IncompatibleConfigException(
+                f"Incompatible storage format in {Configuration.FILEPATH}. "
+                "Please remove the config and try again."
+            ) from e
```
chaudum: :tada:

croud/logout.py
```diff
@@ -30,12 +30,15 @@ def logout(args: Namespace) -> None:
...
 
+    loop.run_until_complete(make_request(env, token))
     Configuration.set_token("")
+
     print_info("You have been logged out.")
 
 
-async def make_request() -> None:
-    async with HttpSession() as session:
+async def make_request(env, token) -> None:
```
chaudum: Add typing information:

```suggestion
async def make_request(env: str, token: str) -> None:
```

croud/config.py
```diff
@@ -36,6 +43,24 @@ class Configuration:
...
+                    "current_context": str,
+                    "contexts": {"prod": {"token": str}, "dev": {"token": str}},
+                }
+            }
+        )
+        try:
+            return schema.validate(config)
+        except SchemaError as e:
+            raise IncompatibleConfigException(
+                f"Incompatible storage format in {Configuration.FILEPATH}. "
```
MarkusH: Why don't we handle this gracefully? e.g. "It appears you have an existing configuration file under <path> which is incompatible with the current version of croud. croud can delete the configuration file for you such that you can login again. Do you want croud to delete the configuration file? [Y]es / [N]o: "

mikethebeer: Good suggestion. I just kept it simple for now but this would indeed bring some nice user value.

PR #22, opened 2018-12-05
WIP Added projects list command

alijared:
This will list projects for the currently logged in user, with a specified region


croud/util.py
```diff
@@ -76,3 +83,38 @@ def open_page_in_browser(url: str) -> int:
...
+    token = Configuration.get_token()
+    region = args.region
+
+    loop = asyncio.get_event_loop()
+    rows = loop.run_until_complete(fetch_data())
+
+    if rows:
+        if isinstance(rows, dict):
+            data = rows[data_key]
+            if "data" in data:
```
mikethebeer: could you probably add a comment here to explain why you check if "data" key exists in the dict:

```
me {
   username
   ...
}

vs.

allProjects {
   data {
      name
      region
      ...
   }
}
```

croud/util.py
```diff
@@ -76,3 +83,34 @@ def open_page_in_browser(url: str) -> int:
...
             pass
     return webbrowser.open_new_tab(url)
+
+
+def get_entity_list(query: str, args: Namespace, data_key: str) -> None:
+    if args.env is not None:
+        Configuration.override_context(args.env)
+
+    async def fetch_data(region: str) -> JsonDict:
+        async with HttpSession(region) as session:
```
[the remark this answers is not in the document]
mikethebeer: 👉 https://github.com/crate/croud/blob/master/croud/session.py#L37

croud/util.py
```diff
@@ -76,3 +83,40 @@ def open_page_in_browser(url: str) -> int:
...
             pass
     return webbrowser.open_new_tab(url)
+
+
+def get_entity_list(query: str, args: Namespace, data_key: str) -> None:
+    if args.env is not None:
+        Configuration.override_context(args.env)
+
+    async def fetch_data() -> JsonDict:
+        async with HttpSession(env, token, region) as session:
```
mikethebeer: don't you get warnings/error that variables are used before they are defined since `env`,`token` and `region` are defined later in the code.

croud/util.py
```diff
@@ -76,3 +83,40 @@ def open_page_in_browser(url: str) -> int:
...
+    region = args.region
+
+    loop = asyncio.get_event_loop()
+    rows = loop.run_until_complete(fetch_data())
+
+    if rows:
+        if isinstance(rows, dict):
+            data = rows[data_key]
+            # some queries have two levels to access data
+            # e.g. {allProjects: {data: [{projects}]}} vs. {me: {data})
```
mikethebeer: 👍

croud/util.py
alijared: Nope

PR #24, opened 2018-12-10
Added clusters subcommand

alijared:
This lists clusters for a current user, in a specified region, and can be filtered by project ID

### DIFF
--- a/croud/clusters/list.py
+++ b/croud/clusters/list.py
@@ -38,6 +38,7 @@ def clusters_list(args: Namespace) -> None:
             crateVersion
             projectId
             username
+            fqdn
         }
     }
 }
--- a/croud/clusters/list.py
+++ b/croud/clusters/list.py
@@ -1,0 +1,51 @@
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.  See the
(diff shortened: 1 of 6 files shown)


croud/clusters/list.py
```diff
@@ -0,0 +1,51 @@
...
+            hardwareSpecs
+            crateVersion
+            projectId
+            username
+        }
+    }
+}
+    """
+
+    if args.project_id is not None:
```
mikethebeer: please add a test that tests if the query gets built correctly.

tests/unit_tests/test_commands.py
```diff
@@ -79,3 +80,59 @@ def test_logout(self, mock_load_config, mock_print_info, mock_set_token):
...
+        )
+        clusters_list(args)
+        get_entity_list.assert_called_once_with(query, args, "allClusters")
+
+    @mock.patch("croud.clusters.list.get_entity_list")
+    def test_list_clusters_pid(self, get_entity_list):
+        query = (
+            """
+{
+    allClusters (filter: %s) {
```
mikethebeer: work with `str.format()` here instead of `%`

tests/unit_tests/test_commands.py
```diff
@@ -79,3 +80,59 @@ def test_logout(self, mock_load_config, mock_print_info, mock_set_token):
...
+            numNodes
+            hardwareSpecs
+            crateVersion
+            projectId
+            username
+        }
+    }
+}
+    """
+            % '{by: PROJECT_ID, op: EQ, value: "60d398b4-455b-49dc-bfe9-04edf5bd3eb2"}'
```
mikethebeer: why not use it directly in the query?

mikethebeer: ok.

alijared: Doesn't work either because of the curly braces in the string. That was what I tried first. This is the only thing I could use that would work as expected, as pass tests

mikethebeer:
retest this please

croud/clusters/list.py
```diff
@@ -0,0 +1,52 @@
...
+            projectId
+            username
+            fqdn
+        }
+    }
+}
+    """
+
+    if args.project_id is not None:
+        filter_value: str = '{by: PROJECT_ID, op: EQ, value: "%s"}' % args.project_id
```
MarkusH: No need for the `: str` type annotation. Neither in the line below.

PR #34, opened 2018-12-13
Added config get and set subcommands

alijared:
This adds the ability for users to store default settings (currently env, region and output format). This is so that users do not have to always specify an output format, or region if their most-used format and region is different to our preselected default. This is useful because you must specify a region when querying projects or clusters, for example, and our default is bregenz.a1, but a user might have no clusters there, and therefore always be forced to parse something like `--region eastus.azure`. The `get` command simple returns a stored setting, for example `croud config get region` will print the currently stored default region. To change this you can run `croud config set --region bregenz.a1`. Setting multiple configuration settings at once is supported, e.g. `croud config set --region eastus.azure --output-fmt table`


alijared:
@mikethebeer the methods used by get and set are already tested. I will write specific tests anyway

mikethebeer:
@alijared I mean those 2 https://github.com/crate/croud/pull/34/files#diff-469b3e03981d2fc872c640304756fd02R171 for testing

PR #36, opened 2018-12-14
Fixed unit tests so that they do not interact with the same variable reference

alijared:
The unit tests were referencing the same variable while testing commands (configuration class instance), so after running the login command, the auth context on the shared configuration class instance was updated. This PR fixes that by mocking the `set_context` method, and using a function to return default configuration settings for the mocked `load_config` return value

### DIFF
--- a/tests/unit_tests/test_commands.py
+++ b/tests/unit_tests/test_commands.py
@@ -28,20 +28,10 @@ from croud.logout import logout
 from croud.server import Server
 
 
-def default_config() -> dict:
-    return {
-        "auth": {
-            "current_context": "prod",
-            "contexts": {"prod": {"token": ""}, "dev": {"token": ""}},
-        },
-        "region": "bregenz.a1",
-        "output_fmt": "table",
-    }
-
-
 class TestLogin(unittest.TestCase):
+    @mock.patch("croud.config.Configuration.set_context")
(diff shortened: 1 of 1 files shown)


tests/unit_tests/test_commands.py
```diff
@@ -28,42 +28,20 @@
...
-
-    @mock.patch("croud.config.write_config")
-    @mock.patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
-    def test_set_top_level_setting(self, mock_load_config, mock_write_config):
-        config = Configuration.DEFAULT_CONFIG
-        config["region"] = "eastus.azure"
-
-        config_set(Namespace(region="eastus.azure"))
-        mock_write_config.assert_called_once_with(config)
+def default_config() -> dict:
```
mikethebeer: since you mocked `override_context` now (which unknowingly changed the auth context) isn't it possible to use `Configuration.DEFAULT_CONFIG` again?

alijared: Unfortunately not no. Not sure why yet, I'm looking into it

alijared: Fixed. Need to override `set_context` as well

mikethebeer: Great. Thx for the update.

PR #37, opened 2019-01-07
Fix: Login page needs to be picked according to the env set in current_context.

mikethebeer:
This fixes a problem that occurred if `current_context: dev` the login has been opened on a prod environment instead of dev.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Fix: Login page needs to be picked according to the env set in current_context.
+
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Fix: Login page needs to be picked occording to the env set in current_context.
```
alijared: Please fix typo

alijared: @mikethebeer it's 'according', not 'occording'

PR #40, opened 2019-01-09
Added organizations create subcommand

alijared:
This command is only useable by super users

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -24,11 +24,19 @@
 import colorama
 
 from croud.clusters.list import clusters_list
-from croud.cmd import CMD, output_fmt_arg, project_id_arg, region_arg
+from croud.cmd import (
+    CMD,
+    org_name_arg,
+    org_plan_type_arg,
+    output_fmt_arg,
+    project_id_arg,
+    region_arg,
+)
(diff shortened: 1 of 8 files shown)


croud/__main__.py
```diff
@@ -39,7 +47,7 @@ def main():
     commands: dict = {
         "me": {
             "description": "Prints the current logged in user",
-            "extra_args": [output_fmt_arg, region_arg],
+            "extra_args": [output_fmt_arg],
```
mikethebeer: I think this change has nothing to do with this feature. Please make a separate PR if needed.

tests/unit_tests/test_util.py
```diff
@@ -18,17 +18,21 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import unittest
+from argparse import Namespace
```
mikethebeer: where is it used?

alijared: https://github.com/crate/croud/pull/40/files#diff-7d5dce122957336cd058de719c4ed7f9R96

https://github.com/crate/croud/pull/40/files#diff-7d5dce122957336cd058de719c4ed7f9R102

PR #42, opened 2019-01-11
list organizations

lukasbals:


croud/organizations/list.py
```diff
@@ -0,0 +1,47 @@
...
+def organizations_list(args: Namespace) -> None:
+    """
+    Lists organizations
+    """
+
+    query = """
+{
+    allOrganizations {
+        data {
+            name,
```
alijared: The Id field should also be returned

PR #45, opened 2019-01-15
Added users roles add sub command

alijared:
This command is only usable by super users

### DIFF
--- a/croud/cmd.py
+++ b/croud/cmd.py
@@ -158,6 +158,18 @@ def org_plan_type_arg(parser: ArgumentParser) -> None:
     )
 
 
+def resource_id_arg(parser: ArgumentParser, required: bool) -> None:
+    parser.add_argument("--resource", type=str, help="Resource ID", required=required)
+
+
+def role_fqn_arg(parser: ArgumentParser) -> None:
+    parser.add_argument("--role", type=str, help="Role FQN", required=True)
+
+
+def user_id_arg(parser: ArgumentParser, required: bool) -> None:
+    parser.add_argument("--user", type=str, help="User ID", required=required)
+
+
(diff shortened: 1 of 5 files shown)


croud/cmd.py
```diff
@@ -158,6 +158,18 @@ def org_plan_type_arg(parser: ArgumentParser) -> None:
     )
 
 
+def resource_id_arg(parser: ArgumentParser, required: bool) -> None:
+    parser.add_argument("--resource", type=str, help="Resource ID", required=required)
+
+
+def role_fqn_arg(parser: ArgumentParser) -> None:
+    parser.add_argument("--role", type=str, help="Role FQN", required=True)
```
mikethebeer: how is the user informed about the supported FQNs he could use?

alijared: There is a GraphQL query for listing all roles, that has an FQN field. I will add the this query to croud as well

mikethebeer: @alijared ok - I'm fine with that. However, can we link then somehow in the `help` to use the `list roles` command to get a list of all supported FQNs ?

alijared: @mikethebeer I added it to the help text for the command

croud/cmd.py
```diff
@@ -158,6 +158,23 @@ def org_plan_type_arg(parser: ArgumentParser) -> None:
...
 
+def resource_id_arg(parser: ArgumentParser, required: bool) -> None:
+    parser.add_argument("--resource", type=str, help="Resource ID", required=required)
+
+
+def role_fqn_arg(parser: ArgumentParser) -> None:
+    parser.add_argument(
+        "--role",
+        type=str,
+        help="Role FQN. Run `croud roles list` for a list of available roles",
```
mikethebeer: 👍

PR #47, opened 2019-01-16
implemented croud users roles list command

lukasbals:


croud/__main__.py
```diff
@@ -129,7 +130,11 @@ def main():
                                 role_fqn_arg,
                             ],
                             "calls": roles_add,
-                        }
+                        },
+                        "list": {
+                            "description": "Lists all roles of a user",
+                            "calls": roles_list,
```
alijared: Can you add the output format argument so users can choose how the list is printed, please?

lukasbals: ohh ... thanks 👍

PR #52, opened 2019-01-22
Added eastus2 to available regions.

christianbader:


croud/cmd.py
```diff
@@ -201,7 +201,7 @@ def region_arg(parser: ArgumentParser) -> None:
     parser.add_argument(
         "-r",
         "--region",
-        choices=["westeurope.azure", "eastus.azure", "bregenz.a1"],
+        choices=["westeurope.azure", "eastus.azure", "eastus2.azure" "bregenz.a1"],
```
mikethebeer: missing comma

```suggestion
        choices=["westeurope.azure", "eastus.azure", "eastus2.azure", "bregenz.a1"],
```

quodt: Shouldn't this info be provided by the backend?

quodt: No test changes required?

alijared: This would make running commands super slow, because we'd have to wait for the backend to reply with the options lists even just ``croud -h``, and whenever we run a command, the choices will need to be populated, then the parser can parse the options/validate them. Unless during the installation process a curl request to a public gql endpoint that has this list is made, and updates the code via templating... But this is messy

PR #54, opened 2019-01-22
error message from graphql if a user haves no permission

lukasbals:


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Fix: Print the error message from graphql
```
mikethebeer: I would write it like this (suggestion) @alijared what do u think?

```suggestion
- Fix: Delegate all occurring error messages to the console output 
```

alijared: lgtm

PR #56, opened 2019-01-22
copy edit

mechanomi:
here's my final copy edit of the docs

please pay special attention to the "Commands" document. I've changed a few things around based on my understanding, but there's the possibility I've made a mistake

I have added an example for ``users roles add``. can you please take a look at that? I would like the example to mirror the sorts of values that would be used in a real-world example. can you provide feedback on that?

### DIFF
--- a/docs/docs/commands.txt
+++ b/docs/docs/commands.txt
@@ -1,75 +1,87 @@
+.. _commands:
+
 ========
 Commands
 ========
 
-Croud has many commands that are able to interact with various CrateDB Cloud resources.
+You can supply commands to the :ref:`Croud CLI <getting-started>` tool, like
+so:
 
-.. NOTE::
-    Most commands are split by the resource that they interact with.
+.. code-block:: console
+
+    sh$ croud [COMMAND] [OPTIONS]
+
+For example:
+
+.. code-block:: console
+
+    sh$ croud login
+
+You can get the list of available commands, like so:
+
+.. code-block:: console
+
+    sh$ croud --help
+
+To get help for a specific command, you can append ``--help``:
+
+.. code-block:: console
+
+    sh$ croud login --help
+
+.. All commands support the ``--env`` flag, which can be used to select between
+.. the ``prod`` and ``dev`` environments. This is an internal feature used by
+.. developers of the Croud CLI tool itself.
 
 .. rubric:: Table of Contents
 
 .. contents::
    :local:
 
-Command-Line Options
-====================
-
-+------------------------+--------------------------------------------------+
-| Argument               | Description                                      |
-+------------------------+--------------------------------------------------+
-| ``-h``,                | Print the help message, then exits.              |
-| ``--help``             |                                                  |
-+------------------------+--------------------------------------------------+
-| ``-v``,                | Prints the program's version number, then exits. |
-| ``--version``          |                                                  |
-+------------------------+--------------------------------------------------+
-
-Sub-Commands
-============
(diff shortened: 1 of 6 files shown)


docs/docs/commands.txt
```diff
@@ -239,48 +267,56 @@ Lists all organizations that a user is part of.
...
+|                           |          | - ``bregenz.a1``           |
 |                           |          | - ``westeurope.azure``     |
 |                           |          | - ``eastus.azure``         |
 +---------------------------+----------+----------------------------+
 | ``--output-fmt <STRING>`` | No       | The desired output format. |
 |                           |          |                            |
-|                           |          | Options:                   |
+|                           |          | One of:                    |
 |                           |          |                            |
 |                           |          | - ``json`` (default)       |
```
alijared: JSON isn't a default, because the defaults will change if they run ``croud config set --output-fmt json|table``

docs/docs/commands.txt
```diff
@@ -358,32 +405,50 @@ of. The list is filterable by project ID.
...
+
+You must specify a second subcommand, like so:
+
+.. code-block:: console
+
+    sh$ croud users roles [SUBCOMMAND] [OPTIONS]
 
 .. users.roles.add:
 
 ``add``
```
lukas-bernert: We now also have a "remove" command. We should add it to the documentation as well. `croud users roles remove --user userid --resource resourceid --role role`

docs/docs/commands.txt
```diff
@@ -393,28 +458,29 @@ For example:
...
 +---------------------------+----------+------------------------------+
 
-Assigns a new role to a user. **Note**:
-this command is only usable by users with
-super user privileges
+For example:
+
+.. code-block:: console
+
+    sh$ croud users roles add --user 'alice' --resource '14569834' --role 'admin'
```
alijared: The ``--user`` option is the users ID, and the ``--role`` option is a role FQN, an example being ``org_admin``

docs/docs/commands.txt
```diff
@@ -1,75 +1,87 @@
...
 
-+------------------------+--------------------------------------------------+
-| Argument               | Description                                      |
-+------------------------+--------------------------------------------------+
-| ``-h``,                | Print the help message, then exits.              |
-| ``--help``             |                                                  |
-+------------------------+--------------------------------------------------+
-| ``-v``,                | Prints the program's version number, then exits. |
-| ``--version``          |                                                  |
-+------------------------+--------------------------------------------------+
```
alijared: @lukas-bernert requested this table exists, like on Crash, why did you remove it? Can you put it back please?

mechanomi: the Crash docs have three documents:

- getting started
- running Crash
- commands

at the moment, Croud only has two:

- getting started
- commands

the `--version` and `--help` documentation lives in the "running Foo" section. because it is neither a command, nor is it covered by the getting started instructions

I considered creating a new document, "running Croud", but the document would literally only have this two-row table at the moment because the tool is still so simple. that seemed silly to me

so what I've done, to address your comment here, is add a "Command-Line Options" section to the getting started document

I did it this way because the commands doc is really only concerning itself with the commands you can run. and I would have to change the structure and even the title and file name to make it fit this table. but adding it to the getting started docs was as easy as adding a new section at the end

as this tool grows in complexity, we can look at creating a separate "running Croud" that just documents how to run it from the command line (environment variables, return values, and so on)

PR #61, opened 2019-01-29
Added users list sub command

alijared:
This adds the ``croud users list`` sub command, that allows users to list all users within organizations that they are part of. Super users are able to filter these by organization ID (``--org-id``), or list all users who are not part of any organization (``--no-org``)

### DIFF
--- a/tests/unit_tests/test_commands.py
+++ b/tests/unit_tests/test_commands.py
@@ -28,6 +28,7 @@ from croud.logout import logout
 from croud.organizations.create import organizations_create
 from croud.organizations.list import organizations_list
 from croud.server import Server
+from croud.users.list import users_list
 from croud.users.roles.add import roles_add
 from croud.users.roles.list import roles_list
 from croud.users.roles.remove import roles_remove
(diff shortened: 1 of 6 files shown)


tests/unit_tests/test_commands.py
```diff
@@ -398,3 +399,83 @@ def test_users_roles_remove(self, mock_gql_mutation, mock_print_format):
         )
         roles_remove(args)
         mock_print_format.assert_called_once_with(self.authd_response, "json")
+
+
+class TestUsersList(unittest.TestCase):
+    @mock.patch("croud.users.list.get_entity_list")
```
mikethebeer: Would it makes sense to use [@pytest.mark.parameterize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) here? There is a lot of boilerplate code in these tests, so I think it might help to reduce this.

alijared: I don't really think so, but I can add it if you really think it'll be useful. It's not going to reduce that much of the boilerplate code though, and the annotations, in my opinion, aren't as readable

mikethebeer: Ok, maybe as a clean-up task.

PR #64, opened 2019-02-04
Created Query class for handling GraphQL queries

alijared:
This PR refactors a lot of code to use a new query class that helps with sending queries to GraphQL. The reason being that there were a few messy function in utils which I felt didn't belong there, and still meant there was a bit of replicated code, especially with mutations. Now, queries can be sent and printed using the Query class


tests/unit_tests/test_gql.py
```diff
@@ -0,0 +1,158 @@
...
+from croud.config import Configuration
+from croud.exceptions import GQLError
+from croud.gql import Query
+
+
+class TestHttpSession(unittest.TestCase):
+    _data = {"data": {"key": "value"}}
+    _error_data = {"errors": [{"message": "error"}]}
+
+    @mock.patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
```
mikethebeer: probably u should use the [@pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) annotation to test different inputs. Feels much cleaner to me.

tests/unit_tests/test_gql.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.assertEqual("dev", query._env)
+
+        args = Namespace(env=None)
+        query = Query("{}", args)
+
+        self.assertEqual("prod", query._env)
+
+    @mock.patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
+    def test_output_fmt_set(self, mock_load_config):
```
mikethebeer: [@pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) annotation (see :point_up: )

tests/unit_tests/test_gql.py
```diff
@@ -0,0 +1,158 @@
...
+
+        self.assertEqual("table", query._output_fmt)
+
+        args = Namespace(env="dev")
+        query = Query("{}", args)
+
+        self.assertEqual("table", query._output_fmt)
+
+    @mock.patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
+    def test_region_set(self, mock_load_config):
```
mikethebeer: [@pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) annotation (see :point_up: )

PR #82, opened 2019-02-13
added graphql variables

Taliik:
This PR adds the support for sending GraphQL requests using variables which simplifies sending input arguments for mutations.

The PR contains of 3 commits:
- https://github.com/crate/croud/pull/82/commits/a32e0770ea0e2657da2b88fc3e7db6fe030d8fa6 : enhance testability of GraphQL client
- https://github.com/crate/croud/pull/82/commits/8cecccb1e1ce307b724b5014f8f4314cff9ea015 : introduce variables for GraphQL requests
- https://github.com/crate/croud/pull/82/commits/8326c46ed3a02dbd3c353cb589f2a6317e8dc7e4 : use variables for existing mutations that require input arguments

### DIFF
--- a/croud/gql.py
+++ b/croud/gql.py
@@ -34,43 +34,32 @@ class Query:
         self._query = query
         self._token = Configuration.get_token()
 
-        self._set_env(args.env)
-        self._set_output_fmt(args)
-        self._set_region(args)
+        self._env = args.env or Configuration.get_env()
+        self._output_fmt = (
+            hasattr(args, "output_fmt")  # certain commands do not have an output format
+            and args.output_fmt
+            or Configuration.get_setting("output_fmt")
+        )
+        self._region = (
+            hasattr(args, "region")  # certain commands do not have region
+            and args.region
+            or Configuration.get_setting("region")
+        )
         self._endpoint = endpoint
 
         self._error: Optional[str] = None
         self._response: Optional[JsonDict] = None
 
-    def _set_env(self, env: Optional[str]):
-        self._env = env or Configuration.get_env()
-
-    def _set_output_fmt(self, args: Namespace):
-        if hasattr(args, "output_fmt") and args.output_fmt is not None:
-            fmt = args.output_fmt
-        else:
-            fmt = Configuration.get_setting("output_fmt")
-
-        self._output_fmt = fmt
-
-    def _set_region(self, args: Namespace):
-        if hasattr(args, "region") and args.region is not None:
-            region = args.region
-        else:
-            region = Configuration.get_setting("region")
-
-        self._region = region
-
-    async def _fetch_data(self, variables: Optional[Dict]) -> JsonDict:
+    async def _fetch_data(self, body: str, variables: Optional[Dict]) -> JsonDict:
         async with HttpSession(self._env, self._token, self._region) as session:
-            return await session.fetch(self._query, variables, endpoint=self._endpoint)
+            return await session.fetch(body, variables, endpoint=self._endpoint)
 
-    def run(self, variables: Optional[Dict]) -> JsonDict:
+    def run(self, body: str, variables: Optional[Dict]) -> JsonDict:
         loop = asyncio.get_event_loop()
-        return loop.run_until_complete(self._fetch_data(variables))
+        return loop.run_until_complete(self._fetch_data(body, variables))
 
     def execute(self, variables: Dict = None):
-        response = self.run(variables)
+        response = self.run(self._query, variables)
 
         if "errors" in response:
             self._response = None
--- a/croud/gql.py
+++ b/croud/gql.py
@@ -21,10 +21,9 @@
 
 import asyncio
 from argparse import Namespace
-from typing import Optional
+from typing import Dict, Optional
 
 from croud.config import Configuration
-from croud.exceptions import GQLError
 from croud.printer import print_error, print_format, print_info
 from croud.session import DEFAULT_ENDPOINT, HttpSession
 from croud.typing import JsonDict
@@ -40,8 +39,8 @@ class Query:
         self._set_region(args)
         self._endpoint = endpoint
 
-        self._error = None
-        self._response = None
+        self._error: Optional[str] = None
+        self._response: Optional[JsonDict] = None
 
     def _set_env(self, env: Optional[str]):
         self._env = env or Configuration.get_env()
@@ -62,29 +61,26 @@ class Query:
 
         self._region = region
 
-    async def _fetch_data(self) -> JsonDict:
+    async def _fetch_data(self, variables: Optional[Dict]) -> JsonDict:
         async with HttpSession(self._env, self._token, self._region) as session:
-            return await session.fetch(self._query, endpoint=self._endpoint)
+            return await session.fetch(self._query, variables, endpoint=self._endpoint)
 
-    def _get_rows(self) -> JsonDict:
-        data = self.run()
-
-        if "errors" in data:
-            raise GQLError(data["errors"][0]["message"])
-        return data["data"]
-
-    def run(self) -> JsonDict:
+    def run(self, variables: Optional[Dict]) -> JsonDict:
         loop = asyncio.get_event_loop()
+        return loop.run_until_complete(self._fetch_data(variables))
 
-        return loop.run_until_complete(self._fetch_data())
+    def execute(self, variables: Dict = None):
+        response = self.run(variables)
 
-    def execute(self):
-        try:
-            self._response = self._get_rows()
-            self._error = None
-        except GQLError as e:
-            self._error = str(e)
+        if "errors" in response:
             self._response = None
+            self._error = response["errors"][0]["message"]
+        elif "data" in response:
+            self._response = response["data"]
+            self._error = None
+        else:
+            self._response = None
+            self._error = None
 
 
 def print_query(query: Query, key: str = None):
--- a/croud/session.py
+++ b/croud/session.py
@@ -60,9 +60,13 @@ class HttpSession:
             cookies={"session": self.token}, connector=conn, headers=headers
         )
 
-    async def fetch(self, query: str, endpoint=DEFAULT_ENDPOINT) -> JsonDict:
+    async def fetch(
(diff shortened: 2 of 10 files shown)


croud/gql.py
```diff
@@ -62,25 +62,25 @@ def _set_region(self, args: Namespace):
...
 
-    def run(self) -> JsonDict:
+    def run(self, variables: Dict) -> JsonDict:
         loop = asyncio.get_event_loop()
 
-        return loop.run_until_complete(self._fetch_data())
+        return loop.run_until_complete(self._fetch_data(variables))
 
-    def execute(self):
+    def execute(self, variables={}):
```
MarkusH: Please don't define mutable objects in function signatures. They can lead to_really interesting_ behaviors:

```python
In [1]: def mutate(d): 
   ...:     d["x"] = "y" 
   ...:                                                                                                                                                                                           

In [2]: def foo(should_mutate, a={}): 
   ...:     print(a) 
   ...:     if should_mutate: 
   ...:         mutate(a) 
   ...:     print(a) 
   ...:                                                                                                                                                                                           

In [3]: foo(False)                                                                                                                                                                                
{}
{}

In [4]: foo(True)                                                                                                                                                                                 
{}
{'x': 'y'}

In [5]: foo(False)                                                                                                                                                                                
{'x': 'y'}
{'x': 'y'}
```

croud/gql.py
```diff
@@ -62,25 +62,25 @@ def _set_region(self, args: Namespace):
 
         self._region = region
 
-    async def _fetch_data(self) -> JsonDict:
+    async def _fetch_data(self, variables: Dict) -> JsonDict:
```
MarkusH: I think `variables` should be `Optional[Dict[str, Any]]`

croud/gql.py
```diff
@@ -62,25 +62,25 @@ def _set_region(self, args: Namespace):
...
         async with HttpSession(self._env, self._token, self._region) as session:
-            return await session.fetch(self._query, endpoint=self._endpoint)
+            return await session.fetch(self._query, variables, endpoint=self._endpoint)
 
-    def _get_rows(self) -> JsonDict:
-        data = self.run()
+    def _get_rows(self, variables: Dict) -> JsonDict:
```
MarkusH: I think `variables` should be `Optional[Dict[str, Any]]`

croud/gql.py
```diff
@@ -62,25 +62,25 @@ def _set_region(self, args: Namespace):
...
+        data = self.run(variables)
 
         if "errors" in data:
             raise GQLError(data["errors"][0]["message"])
         return data["data"]
 
-    def run(self) -> JsonDict:
+    def run(self, variables: Dict) -> JsonDict:
```
MarkusH: I think `variables` should be `Optional[Dict[str, Any]]`

croud/session.py
```diff
@@ -60,9 +60,13 @@ def __init__(
             cookies={"session": self.token}, connector=conn, headers=headers
         )
 
-    async def fetch(self, query: str, endpoint=DEFAULT_ENDPOINT) -> JsonDict:
+    async def fetch(
+        self, query: str, variables: Dict, endpoint=DEFAULT_ENDPOINT
```
MarkusH: I think `variables` should be `Optional[Dict[str, Any]]`

croud/gql.py
chaudum: These `Optional` type annotations lead to strange analysis errors :(

chaudum: This would work:

```diff
diff --git a/croud/gql.py b/croud/gql.py
index 6ae30b3..f1d8d84 100644
--- a/croud/gql.py
+++ b/croud/gql.py
@@ -40,8 +40,8 @@ class Query:
         self._set_region(args)
         self._endpoint = endpoint
 
-        self._error = None
-        self._response = None
+        self._error: Optional[str] = None
+        self._response: Optional[JsonDict] = None
 
     def _set_env(self, env: Optional[str]):
         self._env = env or Configuration.get_env()
@@ -62,23 +62,22 @@ class Query:
 
         self._region = region
 
-    async def _fetch_data(self, variables: Dict) -> JsonDict:
+    async def _fetch_data(self, variables: Optional[Dict]) -> JsonDict:
         async with HttpSession(self._env, self._token, self._region) as session:
             return await session.fetch(self._query, variables, endpoint=self._endpoint)
 
-    def _get_rows(self, variables: Dict) -> JsonDict:
+    def _get_rows(self, variables: Optional[Dict]) -> JsonDict:
         data = self.run(variables)
 
         if "errors" in data:
             raise GQLError(data["errors"][0]["message"])
         return data["data"]
 
-    def run(self, variables: Dict) -> JsonDict:
+    def run(self, variables: Optional[Dict]) -> JsonDict:
         loop = asyncio.get_event_loop()
-
         return loop.run_until_complete(self._fetch_data(variables))
 
-    def execute(self, variables={}):
+    def execute(self, variables: Dict = None):
         try:
             self._response = self._get_rows(variables)
             self._error = None
diff --git a/croud/session.py b/croud/session.py
index c91d93c..386698e 100644
--- a/croud/session.py
+++ b/croud/session.py
@@ -61,7 +61,7 @@ class HttpSession:
         )
 
     async def fetch(
-        self, query: str, variables: Dict, endpoint=DEFAULT_ENDPOINT
+        self, query: str, variables: Optional[Dict], endpoint=DEFAULT_ENDPOINT
     ) -> JsonDict:
         url = self.url + endpoint
         resp = await self.client.post(
```

funnily, if self._error and self._response are not annotated, the unrelated `Optional[Dict]` would not work.

croud/gql.py
chaudum: What do you think about merging the `_get_rows` method into the `execute` method?
This would prevent raising and catching the `GQLError` exception, which is not needed.

```py
    def execute(self, variables: Dict = None):
        response = self.run(variables)

        if "errors" in response:
            self._response = None
            self._error = response["errors"][0]["message"]
        elif "data" in response:
            self._response = response["data"]
            self._error = None
        else:
            self._response = None
            self._error = None
```

chaudum: This would imply a lot of test changes I guess.

tests/unit_tests/test_gql.py
```diff
@@ -67,27 +67,26 @@ def test_region_set(mock_load_config, input, expected):
     assert query._region == expected
 
 
+@patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
 class TestGQL(unittest.TestCase):
```
chaudum: I would even patch `print_info` and `print_error` on a class level.

tests/unit_tests/test_gql.py
```diff
@@ -67,27 +67,26 @@ def test_region_set(mock_load_config, input, expected):
...
     _data = {"data": {"key": "value"}}
     _error_data = {"errors": [{"message": "error"}]}
+    _vars = {"productId": "123", "consumerSetId": "123abc"}
 
-    @patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
     @patch.object(Query, "run", return_value=_data)
     def test_get_rows(self, mock_run, mock_load_config):
         query = Query("{}", Namespace(env="dev"))
-        data = query._get_rows()
+        data = query._get_rows(variables={})
```
chaudum: can be removed if variables is optional

tests/unit_tests/test_gql.py
```diff
@@ -67,27 +67,26 @@ def test_region_set(mock_load_config, input, expected):
...
         self.assertEqual(self._data["data"], data)
 
-    @patch("croud.config.load_config", return_value=Configuration.DEFAULT_CONFIG)
     @patch.object(Query, "run", return_value=_error_data)
     def test_get_rows_error(self, mock_run, mock_load_config):
         query = Query("{}", Namespace(env="dev"))
 
         with self.assertRaises(GQLError):
-            query._get_rows()
+            query._get_rows(variables={})
```
chaudum: same here

tests/unit_tests/test_gql.py
```diff
@@ -150,3 +144,9 @@ def test_print_result_empty(self, mock_get_rows, mock_print_info, mock_load_conf
         print_query(query, "key")
 
         mock_print_info.assert_called_once_with("Result contained no data to print.")
+
+    @patch.object(Query, "run", return_value={"data": []})
+    def test_run_with_variables(self, mock_get_rows, mock_load_config):
+        query = Query("{}", Namespace(env="dev"))
+        query.execute(variables={"productId": "123", "consumerSetId": "123abc"})
```
chaudum: since variables are only used with this single test, you should also only define them with this scope.
a variable can be used for both the invocation of `execute` and the assertion.

tests/unit_tests/test_session.py
```diff
@@ -62,7 +62,7 @@ def test_query_success(self, mock_token, mock_env):
                     conn=connector,
                     headers=headers,
                 ) as session:
-                    result = await session.fetch(me_query)
+                    result = await session.fetch(me_query, variables={})
```
chaudum: Should rather be `None`, no?

```suggestion
                    result = await session.fetch(me_query, variables=None)
```

tests/unit_tests/test_session.py
```diff
@@ -92,7 +91,7 @@ def test_query_unauthorized(self, mock_print_error, mock_token, mock_env):
                         conn=connector,
                         headers=headers,
                     ) as session:
-                        await session.fetch(me_query)
+                        await session.fetch(me_query, variables={})
```
chaudum: same here

croud/gql.py
chaudum: `Dict` == `Dict[str, Any]` :)

chaudum:
@MarkusH @alijared wanna have another look?

chaudum:
No, it’s not a public change.

tests/unit_tests/test_session.py
MarkusH: What about this?

```suggestion
                        await session.fetch(me_query)
```

MarkusH:
Nice work, @Taliik @chaudum

chaudum:
@christianbader please have a look

PR #85, opened 2019-02-18
Fixed addUserToOrganization and removeUserFromOrganization mutations …

m-kharbat:
…call to be compatible with recent cloud-app changes.

### DIFF
--- a/croud/organizations/users/commands.py
+++ b/croud/organizations/users/commands.py
@@ -29,7 +29,7 @@ from croud.gql import Query, print_query
 def org_users_add(args: Namespace):
     mutation = textwrap.dedent(
         """
-        mutation addUserToOrganization(input: AddUserToOrganizationInput!) {
+        mutation addUserToOrganization($input: AddUserToOrganizationInput!) {
           addUserToOrganization(input: $input) {
             user {
               uid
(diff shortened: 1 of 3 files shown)


croud/organizations/users/commands.py
```diff
@@ -29,7 +29,7 @@
 def org_users_add(args: Namespace):
     mutation = textwrap.dedent(
         """
-        mutation addUserToOrganization(input: UserInput!) {
+        mutation addUserToOrganization(input: AddUserToOrganizationInput!) {
```
alijared: Can this not just be ``mutation {...``?

chaudum: You need a named query/mutation when using variables :disappointed:

croud/organizations/users/commands.py
```diff
@@ -41,7 +41,13 @@ def org_users_add(args: Namespace):
     """
     ).strip()
 
-    vars = {"user": args.user, "roleFqn": args.role, "organizationId": args.org_id}
+    vars = {"user": args.user}
+    if args.org_id is not None:
+        vars["organizationId"] = args.org_id
+    if args.role is not None:
+        vars["roleFqn"] = args.role
+    vars = {"input": vars}
```
chaudum: What about: 

```py
vars = {
    "input": {
        "organizationId": args.org_id,
        "user": args.user,
        "roleFqn": args.role,
    }
}
```

alijared: No, because then it will pass `None` to the mutation, which isn't a valid value (needs to be `String` or not present at all)

chaudum:
ack

PR #87, opened 2019-02-19
"addRoleToUser" mutation returns success field only

mikethebeer:
Reflects https://github.com/crate/cloud-app/pull/627/commits/50afd53d2ec2ff5799174e4c061e157e305d5a64


mikethebeer:
@m-kharbat I think there is no changelog entry necessary since it has no user facing change.

m-kharbat:
@mikethebeer even if it's a breaking change ?

PR #88, opened 2019-02-19
Improve the wording of some command descriptions

mikethebeer:
This will slightly improve the wording of some command descriptions in
the ``croud projects users`` argument namespace.


lukas-bernert:
lgtm

PR #89, opened 2019-02-19
configure consumer sets

chaudum:


croud/consumersets/commands.py
```diff
@@ -50,12 +47,46 @@ def consumer_sets_list(args: Namespace) -> None:
...
 
     query = Query(body, args, endpoint="/product/graphql")
     query.execute(vars)
     print_query(query, "allConsumerSets")
+
+
+def consumer_sets_edit(args: Namespace) -> None:
+    body = dedent(
+        """
+    mutation ConfigureConsumerSet(id: String!, input: ConfigureConsumerSetInput!) {
```
m-kharbat: just out of curiosity, does this work ? isn't it required to have `$` in variable names: 

`mutation ConfigureConsumerSet(id: String!, input: ConfigureConsumerSetInput!)` > `mutation ConfigureConsumerSet($id: String!, $input: ConfigureConsumerSetInput!)`

alijared: @m-kharbat is right

croud/consumersets/commands.py
```diff
@@ -50,12 +47,46 @@ def consumer_sets_list(args: Namespace) -> None:
     """  # noqa
     ).strip()
 
-    vars = {
+    vars = clean_dict({
```
MarkusH: No need here. The server will treat empty and NULL as not provided.

chaudum: I agree, it's not necessary, but it's safer. And it would be needed if arguments have default values.

PR #91, opened 2019-02-19
WIP Refactored addUserToProject mutation to return user object

alijared:


chaudum:
@alijared Does this need to go into the next release?

PR #96, opened 2019-02-21
Added local env

alijared:
This adds a new environment to croud (local), which can be used for running croud against a local instance of cloud-app (via https://github.com/crate/cloud). As the configuration auth contexts have been appended to you will be logged out when using this new version as it will reformat your config file to support the local context.

### DIFF
--- a/croud/login.py
+++ b/croud/login.py
@@ -24,7 +24,10 @@ from typing import Optional
 from croud.config import Configuration
 from croud.printer import print_error, print_info
 from croud.server import Server
+from croud.session import cloud_url
 from croud.util import can_launch_browser, open_page_in_browser
+
+LOGIN_PATH = "/oauth2/login?cli=true"
 
 
 def login(args: Namespace) -> None:
@@ -66,7 +69,4 @@ def _set_login_env(env: Optional[str]) -> str:
 
 
 def _login_url(env: str) -> str:
-    domain = "cratedb.cloud"
-    if env.lower() == "dev":
-        domain = "cratedb-dev.cloud"
-    return f"https://bregenz.a1.{domain}/oauth2/login?cli=true"
+    return cloud_url(env.lower()) + LOGIN_PATH
--- a/croud/logout.py
+++ b/croud/logout.py
@@ -22,7 +22,9 @@ from argparse import Namespace
 
 from croud.config import Configuration
 from croud.printer import print_info
-from croud.session import HttpSession
+from croud.session import HttpSession, cloud_url
+
+LOGOUT_PATH = "/oauth2/logout"
 
 
 def logout(args: Namespace) -> None:
@@ -41,4 +43,8 @@ def logout(args: Namespace) -> None:
 
 async def make_request(env: str, token: str) -> None:
     async with HttpSession(env, token) as session:
-        await session.logout()
+        await session.logout(_logout_url(env))
+
+
+def _logout_url(env: str) -> str:
+    return cloud_url(env) + LOGOUT_PATH
--- a/croud/session.py
+++ b/croud/session.py
@@ -27,6 +27,7 @@ from aiohttp import ClientSession, ContentTypeError, TCPConnector  # type: ignor
 from croud.printer import print_error
 from croud.typing import JsonDict
 
+CLOUD_LOCAL_URL = "http://localhost:8000"
 CLOUD_DEV_DOMAIN = "cratedb-dev.cloud"
 CLOUD_PROD_DOMAIN = "cratedb.cloud"
 DEFAULT_ENDPOINT = "/graphql"
@@ -46,10 +47,7 @@ class HttpSession:
         self.token = token
 
         if not url:
-            host = CLOUD_PROD_DOMAIN
-            if self.env == "dev":
-                host = CLOUD_DEV_DOMAIN
-            url = f"https://{region}.{host}"
+            url = cloud_url(env, region)
 
         self.url = url
         if conn is None:
@@ -82,11 +80,7 @@ class HttpSession:
                 f"Query failed to run by returning code of {resp.status}. {query}"
             )
 
-    async def logout(self):
-        domain = CLOUD_PROD_DOMAIN
-        if self.env == "dev":
-            domain = CLOUD_DEV_DOMAIN
-        url = f"https://bregenz.a1.{domain}/oauth2/logout"
+    async def logout(self, url: str):
         await self.client.get(url)
 
     async def __aenter__(self) -> "HttpSession":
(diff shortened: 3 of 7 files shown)


mikethebeer:
Have you thought about the ability to define the endpoint (host) as well? There might be the case that the default port `8000` for local development will change (or you might need to change it). If nothing gets set, fallback to `localhost:8000`.

TLDR;
Something like this.

```
$ croud config set --env local --host localhost:8000
```

alijared:
@mikethebeer I thought about it, but I left it out because in order for it to run on a different port you'd have to change the nginx config in cloud and a few other things, which would actually be fairly hassle-some, so it seems unlikely anyone is going to do that. @christianbader, @m-kharbat what do you think?

MarkusH:
> @mikethebeer I thought about it, but I left it out because in order for it to run on a different port you'd have to change the nginx config in cloud and a few other things, which would actually be fairly hassle-some, so it seems unlikely anyone is going to do that. @christianbader, @m-kharbat what do you think?

I don't completely agree with that. E.g. you can have one service run outside docker-compose and thus on a different port or IP.

alijared:
> > @mikethebeer I thought about it, but I left it out because in order for it to run on a different port you'd have to change the nginx config in cloud and a few other things, which would actually be fairly hassle-some, so it seems unlikely anyone is going to do that. @christianbader, @m-kharbat what do you think?
> 
> I don't completely agree with that. E.g. you can have one service run outside docker-compose and thus on a different port or IP.

If you run cloud-app locally to use croud you have to run cloud, in which services are only available through 8000. If you run cloud-app against a different port, it's still only accessible at 8000 through the nginx proxy (using extra_hosts in docker-compose). How else can you run this?

CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Added new local env setting for using croud against a local environment.
```
chaudum: I am questioning whether this change should have a changelog entry. Changelog entries are user-facing changes. The `local` env is in my opinion not intended for regular users.

docs/docs/commands.txt
```diff
@@ -30,7 +30,7 @@ To get help for a specific command, you can append ``--help``:
     sh$ croud login --help
 
 .. All commands support the ``--env`` flag, which can be used to select between
-.. the ``prod`` and ``dev`` environments. This is an internal feature used by
+.. the ``prod``, ``dev`` and ``local`` environments. This is an internal feature used by
```
chaudum: Same as for changelog entry applies here as well.

CHANGES.rst
alijared: I added it because we still use croud and so does @dominikmoosbrugger, and @lukas-bernert so it could be helpful to know when the local env was added - it's easier than looking at PR's. I can remove though as I don't have too strong an opinion?

croud/login.py
```diff
@@ -67,6 +67,6 @@ def _set_login_env(env: Optional[str]) -> str:
 
 def _login_url(env: str) -> str:
     domain = "cratedb.cloud"
-    if env.lower() == "dev":
+    if env.lower() in ["dev", "local"]:
         domain = "cratedb-dev.cloud"
```
MarkusH: I'm not convinced `local` should point to `cratedb-dev.cloud`. I think there's more logic required and we might want it to point to `http://localhost:8000/oauth2/login?cli=true` instead.

docs/docs/commands.txt
MarkusH: I can see why we might not want to have that in the docs, and I think I agree with that.

CHANGES.rst
MarkusH: Who's the intended user for the `--env local` feature? Is it one of "us engineers", POs, or customers? If it's "us engineers" I think not including it is fine.

chaudum: IMO, users like @dominikmoosbrugger and @lukas-bernert should never ever be required to use the local env.
/cc @christianbader

alijared: I'll remove :)

croud/session.py
```diff
@@ -83,10 +86,13 @@ def __init__(
             )
 
     async def logout(self):
-        domain = CLOUD_PROD_DOMAIN
-        if self.env == "dev":
-            domain = CLOUD_DEV_DOMAIN
-        url = f"https://bregenz.a1.{domain}/oauth2/logout"
+        if self.env == "local":
+            url = "http://localhost:8000/oauth2/logout"
```
chaudum: The logout path `/oauth2/logout` could be a constant.

croud/login.py
```diff
@@ -66,6 +66,9 @@ def _set_login_env(env: Optional[str]) -> str:
 
 
 def _login_url(env: str) -> str:
+    if env == "local":
+        return "http://localhost:8000/oauth2/login?cli=true"
```
chaudum: The login path `/oauth2/login` could be a constant.

croud/logout.py
```diff
@@ -41,4 +43,15 @@ def logout(args: Namespace) -> None:
...
     async with HttpSession(env, token) as session:
-        await session.logout()
+        await session.logout(_logout_url(env))
+
+
+def _logout_url(env: str) -> str:
+    if env == "local":
+        domain = "http://localhost:8000"
+    else:
+        host = "cratedb.cloud"
```
chaudum: You could re-use CLOUD_PROD_DOMAIN from https://github.com/crate/croud/pull/96/files#diff-fc96a85f2127221b6aa3b3a312e082a9R30

```suggestion
        domain = env == "dev" and CLOUD_DEV_DOMAIN or CLOUD_PROD_DOMAIN
```

croud/session.py
```diff
@@ -102,3 +95,13 @@ def __init__(
         exc_tb: Optional[TracebackType],
     ) -> None:
         await self.close()
+
+
+def _url(env: str, region: str) -> str:
```
chaudum: This function could be used for both generating the login and logout urls as well?

croud/session.py
```diff
@@ -102,3 +95,13 @@ def __init__(
...
+    if env == "local":
+        return "http://localhost:8000"
```
chaudum: You could also add a `CLOUD_LOCAL_DOMAIN` const https://github.com/crate/croud/pull/96/files#diff-fc96a85f2127221b6aa3b3a312e082a9R30

chaudum:
> Shouldn't we document this feature somewhere in DEVELOP.rst?

Not sure. The feature is used for development of the Cloud, not for development of the CLI itself.

croud/logout.py
```diff
@@ -41,4 +43,8 @@ def logout(args: Namespace) -> None:
 
 async def make_request(env: str, token: str) -> None:
     async with HttpSession(env, token) as session:
-        await session.logout()
+        await session.logout(_logout_url(env))
+
+
+def _logout_url(env: str) -> str:
+    return cloud_url(env) + LOGOUT_PATH
```
chaudum: Should we use `env.lower()` here as well?

alijared: The variable can never be lower here so no :)

alijared:
> > Shouldn't we document this feature somewhere in DEVELOP.rst?
> 
> Not sure. The feature is used for development of the Cloud, not for development of the CLI itself.

Exactly, and there's a PR for cloud that documents this: https://github.com/crate/cloud/pull/68

PR #99, opened 2019-02-22
Switch uid with user argment for removing users from organizations

alijared:
This PR makes it so that croud allows you to remove users from organizations via a user's email address or ID


alijared:
> Isn't there a change in the argument helper docs necessary? See https://github.com/crate/croud/blob/master/croud/__main__.py#L270 so that it's using the `user_id_or_email_arg`

You're totally right

PR #102, opened 2019-02-25
Use variables for GraphQL queries and mutations

chaudum:


croud/users/commands.py
```diff
@@ -17,37 +17,40 @@
...
         }
-    }
     """
+    ).strip()
 
-    queryArgs = ""
-    if args.no_org:
-        queryArgs = "(queryArgs: {noOrg: true})"
-    if args.org_id:
-        queryArgs = f'(queryArgs: {{organizationId: "{args.org_id}"}})'
```
chaudum: @alijared You wrote this so that the `--org-id` argument "overrules" the `--no-org` argument. Where there any reasons why you did this implicitly (there is also a test) instead of returning an error (arguments are mutually exclusive)?

alijared: The arguments are meant to be mutually exclusive, and I'm not 100% sure why I did it like this. I'd suggest just returning an error if both of the arguments exist

chaudum: >  I'd suggest just returning an error if both of the arguments exist

OK, will do that!

croud/users/commands.py
```diff
@@ -29,6 +30,11 @@ def users_list(args: Namespace) -> None:
     List all users within organizations that the logged in user is part of
     """
 
+    if args.org_id and args.no_org:
```
MarkusH: Instead of doing this, can we maybe use [`add_mutually_exclusive_group()`](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) instead. Seems like a better option to do the parsing in the parser than in the implementation.

chaudum: I thought about this, however, I don't see how this would be possible with our custom parsing of arguments.

MarkusH: This should do:
```diff
$ git diff | cat 
diff --git a/croud/__main__.py b/croud/__main__.py
index 3b984a2..b89c4e1 100644
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -41,6 +41,7 @@ from croud.cmd import (
     crate_version_arg,
     no_org_arg,
     org_id_arg,
+    org_id_no_org_arg_mutual_exclusive,
     org_name_arg,
     org_plan_type_arg,
     output_fmt_arg,
@@ -286,10 +287,9 @@ command_tree = {
                 "logged in user is part of.",
                 "extra_args": [
                     output_fmt_arg,
-                    lambda req_opt_group, opt_opt_group: org_id_arg(
+                    lambda req_opt_group, opt_opt_group: org_id_no_org_arg_mutual_exclusive(
                         req_opt_group, opt_opt_group, False
                     ),
-                    no_org_arg,
                 ],
                 "calls": users_list,
             },
diff --git a/croud/cmd.py b/croud/cmd.py
index 50014f6..a7b571d 100644
--- a/croud/cmd.py
+++ b/croud/cmd.py
@@ -254,6 +254,15 @@ def no_org_arg(req_args: _ArgumentGroup, opt_args: _ArgumentGroup) -> None:
     )
 
 
+def org_id_no_org_arg_mutual_exclusive(
+    req_args: _ArgumentGroup, opt_args: _ArgumentGroup, required: bool
+) -> None:
+    group = req_args if required else opt_args
+    exclusive = group.add_mutually_exclusive_group()
+    org_id_arg(None, exclusive, False)
+    no_org_arg(None, exclusive)
+
+
 def org_name_arg(req_args: _ArgumentGroup, opt_args: _ArgumentGroup) -> None:
     req_args.add_argument("--name", type=str, help="Organization Name.", required=True)
```

chaudum: It does not make the code nicer, but works.
Simplifying the parsing code should be another task, thb.

croud/users/roles/commands.py
```diff
@@ -17,67 +17,91 @@
...
 
 from croud.gql import Query, print_query
+from croud.util import clean_dict
 
 
-def _get_mutation_input(args: Namespace) -> str:
-    return f'userId: "{args.user}",roleFqn: "{args.role}",resourceId: "{args.resource}"'
-
-
-def roles_add(args: Namespace) -> None:
```
MarkusH: Why did you change the order of `roles_add` and `roles_list`?

tests/unit_tests/test_commands.py
```diff
@@ -167,51 +167,36 @@ def assert_query(mock_print, expected):
...
-        data {
-            id
-            name
-            numNodes
-            crateVersion
-            projectId
-            username
-            fqdn
+    project_id = gen_uuid()
+    expected_body = dedent(
```
MarkusH: Would you mind changing this to `textwrap.dedent()` or change other occurrences to `dedent()`. At least in places that you're touching anyway.

tests/unit_tests/test_commands.py
```diff
@@ -571,106 +560,109 @@ def test_list(self, mock_run, mock_load_config):
...
+    def test_list_filter_conflict(self, mock_run, mock_load_config):
+        expected_vars = None
+        argv = ["croud", "users", "list", "--org-id", self.org_id, "--no-org"]
+        with mock.patch("sys.stderr.write") as stderr:
+            with pytest.raises(SystemExit) as ex_info:
+                self.assertGql(mock_run, argv, self.expected_body, expected_vars)
+            stderr.assert_called_once()
+            assert (
+                stderr.call_args[0][0]
+                == "Argument --no-org: not allowed with argument --org-id\n\n"
```
MarkusH: Can you move that error message into a variable so the tests looks a bit nicer:
```python
msg = "Argument --no-org: ..."
assert stderr.call_args[0][0] == msg
```

tests/unit_tests/test_commands.py
chaudum: Will change that in an additional commit for all usages.

PR #104, opened 2019-02-26
Print user friendly message if query fails

alijared:
This will make sure that a user friendly message is printed when queries fail to run, instead of the stack trace


croud/session.py
```diff
@@ -65,22 +65,21 @@ def __init__(
...
-                f"Query failed to run by returning code of {resp.status}. {query}"
-            )
+
+        if resp.status == 302:  # login redirect
+            print_error("Unauthorized. Use `croud login` to login to CrateDB Cloud.")
+            await self.client.close()
+            exit(1)
+
+        if resp.status != 200:
+            print_info(f"Query failed to run by returning code of {resp.status}.")
```
chaudum: I am not 100% sure whether we should print info or error here.
What's your opinion @alijared ?

croud/session.py
```diff
@@ -65,22 +65,21 @@ def __init__(
...
+            print_info(query)
+            if variables:
+                print_info(str(variables))
```
chaudum: For even nicer output we could json dump the dict:
```suggestion
                print_info(json.dumps(variables, indent=2))
```

tests/unit_tests/util/fake_server.py
```diff
@@ -93,9 +93,9 @@ def __init__(self, loop: asyncio.AbstractEventLoop):
                         }
                     }
                 )
-        # If the session cookie is invalid the OAuth Proxy will serve the
-        # provider sign-in page (status = 200) instead of a 401 Unauthorized
-        return web.Response(status=200, text="OAuth provider sign-in page")
+            resp = {"data": {"message": "Bad request"}}
```
chaudum: Should this be within an if block?

```suggestion
            elif self._get_query_header(request) == "bad":
                resp = {"data": {"message": "Bad request"}}
                return web.json_response(resp, status=400)
            ...
```

tests/unit_tests/test_session.py
```diff
@@ -107,3 +107,43 @@ def test_query_unauthorized(self, mock_print_error, mock_token, mock_env):
...
+                        ]
+                    )
+
+            fake_cloud = FakeCrateDBCloud(loop=loop)
+            info = loop.run_until_complete(fake_cloud.start())
+            resolver = FakeResolver(info, loop=loop)
+            connector = aiohttp.TCPConnector(loop=loop, resolver=resolver, ssl=True)
+
+            # testing-only: tell the graphql endpoint which query is tested
+            headers = {"query": "fakeQuery"}
```
chaudum: Personally I would call it "bad" or "badRequest" (something that hints to HTTP status 400).

croud/session.py
```diff
@@ -65,22 +65,21 @@ def __init__(
     ) -> JsonDict:
         url = self.url + endpoint
         resp = await self.client.post(
-            url, json={"query": query, "variables": variables}
+            url, json={"query": query, "variables": variables}, allow_redirects=False
```
MarkusH: Are you sure this doesn't break the login flow?

croud/session.py
```diff
@@ -65,22 +65,21 @@ def __init__(
...
-                exit(1)
-        else:
-            raise Exception(
-                f"Query failed to run by returning code of {resp.status}. {query}"
-            )
+
+        if resp.status == 302:  # login redirect
+            print_error("Unauthorized. Use `croud login` to login to CrateDB Cloud.")
+            await self.client.close()
+            exit(1)
```
MarkusH: This feels weird. Can we instead raise an exception that we catch on a higher level? Also, please use [`sys.exit()`](https://docs.python.org/3/library/sys.html#sys.exit).

croud/session.py
MarkusH: I'd go with "error". A user expected this command to work. It didn't. That's an error from the perspective of a user.

Also, the query didn't fail because of the status code, but the status code is an error code because the query failed to run:
```suggestion
            print_error(f"Query failed to run and returned an error code {resp.status}.")
```

I think it would then make sense to look if `resp.json()["errors"]` exists and may format them "nicely".

tests/unit_tests/test_session.py
```diff
@@ -107,3 +107,43 @@ def test_query_unauthorized(self, mock_print_error, mock_token, mock_env):
...
+                            mock.call(query),
+                        ]
+                    )
+
+            fake_cloud = FakeCrateDBCloud(loop=loop)
+            info = loop.run_until_complete(fake_cloud.start())
+            resolver = FakeResolver(info, loop=loop)
+            connector = aiohttp.TCPConnector(loop=loop, resolver=resolver, ssl=True)
+
+            # testing-only: tell the graphql endpoint which query is tested
```
MarkusH: I don't understand that comment.

croud/session.py
chaudum: This happens already when the query (response or error) is printed: https://github.com/crate/croud/blob/master/croud/gql.py#L75

croud/session.py
chaudum: Right!

lukasbals: The login still works ...

tests/unit_tests/test_session.py
lukasbals: What would you prefer?

chaudum: Imo the comment can be removed.

tests/unit_tests/test_session.py
```diff
@@ -107,3 +107,38 @@ def test_query_unauthorized(self, mock_print_error, mock_token, mock_env):
...
+                    await session.fetch(query, variables={})
+                    mock_print_info.assert_has_calls([mock.call(query)])
+
+            fake_cloud = FakeCrateDBCloud(loop=loop)
+            info = loop.run_until_complete(fake_cloud.start())
+            resolver = FakeResolver(info, loop=loop)
+            connector = aiohttp.TCPConnector(loop=loop, resolver=resolver, ssl=True)
+
+            # testing-only: tell the graphql endpoint which query is tested
+            headers = {"query": "badRequest"}
```
chaudum: The `query` header is responsible for the fake response, see https://github.com/crate/croud/pull/104/files#diff-1cd72360332543c6203a461d33f4d99fR96
Maybe this could be made more explicit, e.g. by passing it to the `test_query` coroutine and name?

chaudum: @MarkusH ^ This should answer your question.

MarkusH: Ahhhh, :bulb: .

tests/unit_tests/test_session.py
MarkusH:
```suggestion
            # Configure which response the FakeCrateDBCloud instance returns
```

PR #108, opened 2019-03-21
Standardize help texts for commands and arguments

MarkusH:


croud/__main__.py
```diff
@@ -290,7 +293,7 @@
                         "calls": org_users_add,
                     },
                     "remove": {
-                        "help": "Remove user from organization",
+                        "help": "Add the selected user from an organization.",
```
lukas-bernert: "to" instead of "from"?

croud/__main__.py
```diff
@@ -307,8 +310,7 @@
         "help": "Manage CrateDB Cloud users.",
         "sub_commands": {
             "list": {
-                "help": "List all users within organizations that the "
-                "logged in user is part of.",
+                "help": "List all users within the given organizations.",
```
lukas-bernert: sometimes we use "specified" and sometimes "given". Maybe we should agree on one?

croud/__main__.py
```diff
@@ -319,7 +321,7 @@
                 "help": "Manage CrateDB Cloud user roles.",
                 "sub_commands": {
                     "add": {
-                        "help": "Adds a role to a user.",
+                        "help": "Grant the selected user to the chosen role.",
```
lukas-bernert: same here. now we have specified, given, chosen

croud/__main__.py
```diff
@@ -335,7 +337,7 @@
                         "calls": roles_add,
                     },
                     "remove": {
-                        "help": "Removes a role from a user.",
+                        "help": "Revoke the chosen role from the selected user.",
```
lukas-bernert: Can we still call it "remove" instead of revoke? as the command is called "remove" as well?

croud/cmd.py
```diff
@@ -408,7 +428,10 @@ def consumer_eventhub_lease_storage_connection_string_arg(
     group.add_argument(
         "--consumer-eventhub-lease-storage-connection-string",
         type=str,
-        help="Connection string of the lease storage for the EventHub consumer.",
+        help=(
+            "The connection string to a lease storage for the Azure EventHub "
+            "consumer write its lease files to."
```
lukas-bernert: ...lease storage **where?** the Azure EventHub consumer write**s** it's lease files to.. ?

croud/cmd.py
```diff
@@ -432,7 +458,7 @@ def consumer_schema_arg(
     group.add_argument(
         "--consumer-schema",
         type=str,
-        help="Database schema in which to insert.",
+        help="The CrateDB database schema used by the Azure EventHub consumer.",
```
lukas-bernert: above we say "...to use."
and here we say "used by...."

can we stick to one?

croud/__main__.py
```diff
@@ -192,16 +191,18 @@
                 "calls": project_create,
             },
             "list": {
-                "help": "Lists all projects for the current "
-                "user in the specified region.",
+                "help": (
+                    "Lists all projects the current user has access to in "
```
lukas-bernert: Sometimes we use "Lists" sometimes "List". Can we please stick to "List"?

croud/__main__.py
chaudum: Same here:
```suggestion
                        "help": "Unassign a role from the selected user.",
```

croud/cmd.py
```diff
@@ -208,7 +208,7 @@ def region_arg(req_args: _ArgumentGroup, opt_args: _ArgumentGroup) -> None:
         "--region",
         choices=["westeurope.azure", "eastus.azure", "eastus2.azure", "bregenz.a1"],
         type=str,
-        help="Switch region that command will be run on.",
+        help="Switches the region that command will be run on.",
```
chaudum: It does not switch in a sense that it is permanent.
It uses the region for a specific command.

croud/__main__.py
MarkusH: I went with "specified".

croud/__main__.py
MarkusH: Well, "Add the specified role to the specified user" sounds really odd in my ears. Any suggestion?

croud/cmd.py
MarkusH: I hope the rephrasing makes it more clear.

croud/__main__.py
MarkusH: Oops. Completely missed that one :see_no_evil:

croud/__main__.py
```diff
@@ -124,10 +124,10 @@
         },
     },
     "consumer-sets": {
-        "help": "Manage consumer sets for CrateDB Cloud for Azure IoT products.",
+        "help": "Manage consumer sets in CrateDB Cloud.",
```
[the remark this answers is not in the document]
MarkusH: I dropped "CrateDB Cloud" from all commands except login and logout as well as the help for the top-level command.

croud/__main__.py
lukas-bernert: No but what about "Grant the selected user the specified role?"

PR #123, opened 2019-04-16
Update wording on the Azure IoT Hub events endpoint

dominikmoosbrugger:
Include both IoT Hub Events endpoint an Azure EventHub in the description of the connection string and the consumer group.


chaudum:
@dominikmoosbrugger Do you mind merging the PR? You'd need to rebase your branch and update the wording in https://github.com/crate/croud/blob/master/croud/cmd.py#L223 ff

PR #124, opened 2019-04-16
Added REST support

alijared:

As cloud-app is switching to REST, Croud needs to be able to support sending REST requests. This PR adds that ability.

## Checklist


CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Added REST support so that commands can request data from the RESTful, for
```
mikethebeer: I think we should only list user-facing changes

tests/unit_tests/test_rest.py
```diff
@@ -0,0 +1,127 @@
...
+from croud.config import Configuration
+from croud.rest import Client
+from croud.session import HttpSession, RequestMethod
+
+
+@mock.patch.object(Configuration, "get_token", return_value="eyJraWQiOiIx")
+@mock.patch("croud.session.cloud_url", return_value="https://cratedb.local")
+def test_send_success_sets_data_with_key(mock_cloud_url, mock_token):
+    with loop_context() as loop:
+        fake_cloud = FakeCrateDBCloud(loop=loop)
```
mikethebeer: instead of creating and starting the `FakeCrateDBCloud` in every test method I would go with a setup and teardown method in a Test class.

```python
from aiohttp.test_utils import setup_test_loop, teardown_test_loop

class TestRestClient:
    @classmethod
    def setup_class(cls):
        cls.loop = setup_test_loop()
        cls.fake_cloud = FakeCrateDBCloud(loop=cls.loop)
        cls.info = cls.loop.run_until_complete(cls.fake_cloud.start())

    @classmethod
    def teardown_class(cls):
        cls.loop.run_until_complete(cls.fake_cloud.stop())
        teardown_test_loop(cls.loop)

...

PR #130, opened 2019-04-25
Dmo wordingconsumers croud

ghost:
Fix wording of croud for eventhub endpoints.


chaudum:
@dominikmoosbrugger Is this still a valid change?

dominikmoosbrugger:
I think it would be valid, but i question whether we still need to bother if we deprecate the consumer anyhow in favor of Azure Functions/other. No end user will see it and we tend not to use the consumer anymore since architecture guideline suggests otherwise?

croud/cmd.py
```diff
@@ -203,7 +203,7 @@ def consumer_schema_arg(
     group.add_argument(
         "--consumer-schema",
         type=str,
-        help="The CrateDB database schema used by the Azure EventHub consumer.",
+        help="The CrateDB database schema used by the Azure Events endpoint consumer.",
```
chaudum, line 206: Shouldn't it be "Event Hubs" and not "Events" ?
https://docs.microsoft.com/en-us/azure/event-hubs/event-hubs-about

```suggestion
        help="The CrateDB database schema used by the Azure Event Hubs endpoint consumer.",
```

Same for the other help texts as well?

PR #133, opened 2019-04-26
make the `config get` commands respect the output format

chaudum:
Example:

```console
$ croud config get env
+-------+
| env   |
|-------|
| dev   |
+-------+

$ croud config get env -o json
[
  {
    "env": "dev"
  }
]
```

## Checklist

### DIFF
--- a/croud/printer.py
+++ b/croud/printer.py
@@ -26,7 +26,7 @@ from tabulate import tabulate
 from croud.typing import JsonDict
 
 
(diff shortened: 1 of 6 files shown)


croud/printer.py
```diff
@@ -78,7 +78,7 @@ def _tabular(self, rows: Union[List[JsonDict], JsonDict]) -> str:
             # | a   |   b |
             # |-----+-----+
```
MarkusH:
```suggestion
            # |-----+-----|
```

PR #138, opened 2019-05-14
Converted cluster listing and deploying to REST

MarkusH:

Depends on crate/cloud-app#785

## Checklist

### DIFF
--- a/croud/clusters/commands.py
+++ b/croud/clusters/commands.py
@@ -17,13 +17,10 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
-import textwrap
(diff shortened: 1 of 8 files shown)


croud/clusters/commands.py
```diff
@@ -54,35 +51,17 @@ def clusters_deploy(args: Namespace) -> None:
...
+    body = {
+        "crate_version": args.version,
+        "name": args.cluster_name,
+        "password": args.password,
+        "product_name": args.product_name,
+        "product_tier": args.tier,
+        "project_id": args.project_id,
+        "username": args.username,
+    }
+    if args.unit:
```
MarkusH: alternatively, we could make the `unit` parameter to have a `default=0` argument.

PR #139, opened 2019-05-15
Converted consumer listing, deploying, and editing to REST

MarkusH:

Depends on crate/cloud-app#788

The first commit is shared with #138.

## Checklist

### DIFF
--- a/croud/consumers/commands.py
+++ b/croud/consumers/commands.py
@@ -21,6 +21,8 @@ from argparse import Namespace
 from textwrap import dedent
 
 from croud.gql import Query, print_query
+from croud.rest import Client
+from croud.session import RequestMethod
 from croud.util import clean_dict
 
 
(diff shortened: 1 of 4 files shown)


croud/consumers/commands.py
```diff
@@ -68,36 +70,30 @@ def consumers_deploy(args: Namespace) -> None:
...
-            productName
-            productTier
-            instances
-            tableName
-            tableSchema
-        }
-    }
-    """  # noqa
-    ).strip()
+    params = {}
```
chaudum: What is the reasoning behind dropping the usage of `clean_dict()`?

croud/consumers/commands.py
```diff
@@ -87,31 +84,25 @@ def consumers_list(args: Namespace) -> None:
...
-            id
-        }
+    body = {
+        "cluster_id": args.cluster_id,
+        "table_name": args.consumer_table,
+        "table_schema": args.consumer_schema,
     }
-    """  # noqa
-    ).strip()
+    config = {}
```
chaudum: Same here

PR #141, opened 2019-05-20
Add consumers delete command

MarkusH:

- [x] Depends on back-end changes.
- [x] Handle empty responses (HTTP 204) in REST client: #150

## Checklist

### DIFF
--- a/croud/consumers/commands.py
+++ b/croud/consumers/commands.py
@@ -21,6 +21,7 @@ from argparse import Namespace
 
 from croud.rest import Client
 from croud.session import RequestMethod
+from croud.util import require_confirmation
 
 
 def consumers_deploy(args: Namespace) -> None:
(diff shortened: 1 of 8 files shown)


croud/util.py
```diff
@@ -91,3 +91,11 @@ def clean_dict(data: Dict) -> Dict:
         else:
             cleaned[k] = v
     return cleaned
+
+
+def confirm_prompt(msg):
+    msg = f"{msg.rstrip()} [yN] "
+    resp = input(msg).lower()
+    if resp in {"y", "yes"}:
```
chaudum: Cosmetics: the `in` already returns a boolean value, you can write it short as 

```suggestion
    return resp in {"y", "yes"}
```

croud/consumers/commands.py
```diff
@@ -106,3 +108,15 @@ def consumers_edit(args: Namespace) -> None:
         RequestMethod.PATCH, f"/api/v2/consumers/{args.consumer_id}/", body=body
     )
     client.print(keys=["id"])
+
+
+def consumers_delete(args: Namespace) -> None:
+    yes = args.yes
```
chaudum: Since we will need that functionality for other delete actions as well, we could make a decorator that handles the prompt.

```py
@require_confirmation("Are you sure you want to delete the consumer?")
def consumers_delete(args: Namespace) -> None:
    ...
```

PR #148, opened 2019-05-22
Lb/products doc

lukas-bernert:

I am not sure if this is the right place to list our available products but for now I think it should do.

## Checklist


MarkusH:
Before we rush to merge this, how about `croud products list` as a command. It would provide a clear list of available products in the corresponding region. Refs crate/cloud-app#802

docs/commands/clusters.rst
```diff
@@ -68,3 +68,23 @@ Example
...
 .. _string delimitation: https://en.wikipedia.org/wiki/Delimiter
+
+Available Products and Tiers
+----------------------------
+
+
++--------------+----------------+------------------------------------------------+-----------+
+| Product Name | Available Tier | Available Region                               | Disk Type |
++--------------+----------------+------------------------------------------------+-----------+
+| cratedb.az1  | XS             | eastus.azure, eastus2.azure, westeurope.azure  | Premium   |
```
MarkusH: The "Premium" disks are our "Standard" disks. Similarly, we don't have "Ephemeral" disks but "Premium" disks.

docs/commands/clusters.rst
```diff
@@ -68,3 +68,23 @@ Example
...
 
 .. _string delimitation: https://en.wikipedia.org/wiki/Delimiter
+
+Available Products and Tiers
+----------------------------
+
+
++--------------+----------------+------------------------------------------------+-----------+
+| Product Name | Available Tier | Available Region                               | Disk Type |
++--------------+----------------+------------------------------------------------+-----------+
```
MarkusH: I guess you want `Product Name` etc to be table headers, right? In that case, you can use `=` instead of `-` right below as a row separator.
```suggestion
+==============+================+================================================+===========+

```

docs/commands/consumers.rst
```diff
@@ -44,6 +44,20 @@ Example
    | dc40090f-e1df-4974-b751-9fc27c824368 | my-first-consumer | 664e6556-5734-449c-a717-f3533e5f86cf | c95fe9bb-1a95-453d-9edc-30ffe7d8eb03 | eventhub-consumer | s             |           1 | raw__parted | doc           |
    +--------------------------------------+-------------------+--------------------------------------+--------------------------------------+-------------------+---------------+-------------+-------------+---------------+
 
+Available Products and Tiers
+----------------------------
+
++-------------------+----------------+------------------------------------------------+
+| Product Name      | Available Tier | Available Region                               |
++-------------------+----------------+------------------------------------------------+
```
MarkusH: As above
```suggestion
+===================+================+================================================+
```

lukas-bernert:
> Before we rush to merge this, how about `croud products list` as a command. It would provide a clear list of available products in the corresponding region. Refs [crate/cloud-app#802](https://github.com/crate/cloud-app/pull/802)

There's a story for this already: https://trello.com/c/gM2ZLXP1/163-13-list-products-in-cli
As discussed, I'll make sure it gets into the next cycle.

docs/commands/clusters.rst
```diff
@@ -68,3 +68,24 @@ Example
...
 .. _string delimitation: https://en.wikipedia.org/wiki/Delimiter
+
+Available Products and Tiers
+----------------------------
+
+
++--------------+----------------+------------------------------------------------+-----------+
+| Product Name | Available Tier | Available Region                               | Disk Type |
++==============+================+================================================+===========+
+
```
MarkusH: You now have an empty line here that needs to go for proper formatting :-|

lukas-bernert: Thanks

lukas-bernert:
> I don't think this is the right place to document the products. `croud` is a client library, and thus should not know about availability of products, especially, because products and their availability are subject to change.

@chaudum right, I agree. I'd only put it here now until we have a `products list` command in croud, which I'd like to squeeze into the next cycle. But now, by the end of Sprint 3 people will be able to deploy products but there's nowhere they can read about available products. Hence, adding it to the doc for now would be the most minimal viable solution, I guess.

PR #149, opened 2019-05-24
Added croud products list command

m-kharbat:
this command list all available products in the current region.
usage:

    croud products list --kind "cluster" --region "eastus.azure"

## Checklist

### DIFF
--- a/croud/products/commands.py
+++ b/croud/products/commands.py
@@ -1,0 +1,37 @@
+# Licensed to CRATE Technology GmbH ("Crate") under one or more contributor
+# license agreements.  See the NOTICE file distributed with this work for
+# additional information regarding copyright ownership.  Crate licenses
+# this file to you under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.  You may
+# obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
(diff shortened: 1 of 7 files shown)


croud/products/commands.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
```
chaudum: The shebang is not needed, because the file is not executed directly.

docs/commands/products.rst
```diff
@@ -0,0 +1,37 @@
...
+.. argparse::
+   :module: croud.__main__
+   :func: get_parser
+   :prog: croud
+   :path: products
+   :nosubcommands:
+
+
+``products list``
+======================
```
chaudum: Underline should match the amount of characters of the headline

```suggestion
``products list``
=================
```

croud/products/commands.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def products_list(args: Namespace) -> None:
+    """
+    Lists available products
+    """
+
+    client = Client(env=args.env, region=args.region, output_fmt=args.output_fmt)
+    url = "/api/v2/products/"
+    if args.kind:
```
chaudum: The query string should be generated in a secure way, e.g. using urllib

```
>>> qs = parse.urlencode({"kind":"cluster"})
>>> qs
'kind=cluster'
>>> parse.urlunparse(parse.ParseResult(scheme="", netloc="", params="", fragment="", path="/api/v2/products/", query=qs))
'/api/v2/products/?kind=cluster'
```

MarkusH: How about `Client.send(RequestMethod.GET, "/api/v2/prosucts/", params={"kind": kind})` ?

PR #151, opened 2019-05-24
H/session

chaudum:

This PR fixes the problem, that the client did not store the
refreshed access token from the server.
This caused long latencies for API calls in case the access token from
the client was expired, because the server had to refresh it on each
request.

Additional there are two other small cleanup commits.

## Checklist

### DIFF
--- a/croud/rest.py
+++ b/croud/rest.py
@@ -36,10 +36,13 @@ class Client:
     _data: Optional[Union[List[dict], dict]] = None
     _error: Optional[dict] = None
 
-    def __init__(self, env: str = None, region: str = None, output_fmt: str = None):
+    def __init__(
+        self, env: str = None, region: str = None, output_fmt: str = None, loop=None
+    ):
         self._env = env or Configuration.get_env()
         self._region = region or Configuration.get_setting("region")
         self._output_fmt = output_fmt or Configuration.get_setting("output_fmt")
+        self.loop = loop or asyncio.get_event_loop()
 
     def send(
         self,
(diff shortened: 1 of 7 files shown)


croud/rest.py
```diff
@@ -52,18 +52,46 @@ def send(
...
+        params: dict = None,
+    ) -> ClientResponse:
+        async with HttpSession(
+            self._env, Configuration.get_token(), self._region
+        ) as session:
+            return await session.fetch(method, endpoint, body, params)
+
+    async def _decode_response(self, resp: ClientResponse):
+        try:
+            return await resp.json()
```
MarkusH: Why are you using e.g. `return await resp.json()` and not `return resp.json()`? Shouldn't the latter return the awaitable object? Since you're not using the awaited response here I don't see a reason to await it.

croud/rest.py
```diff
@@ -52,18 +52,46 @@ def send(
...
+        self,
+        method: RequestMethod,
+        endpoint: str,
+        body: dict = None,
+        params: dict = None,
+    ) -> ClientResponse:
+        async with HttpSession(
+            self._env, Configuration.get_token(), self._region
+        ) as session:
+            return await session.fetch(method, endpoint, body, params)
```
MarkusH: Can't you just return the awaitable object? See comment below in `_decode_response`.

croud/rest.py
chaudum: `_decode_response` itself is a coroutine that returns a dict.

It you don't `await` the coroutine `resp.json()`, it would never be awaited, and additional the `_decode_response` coroutine would return a coroutine (that then would need to be awaited still).

MarkusH: Thanks for elaborating. asyncio is still this huge pile of mystery to me.

PR #152, opened 2019-05-24
prepare release 0.12.2

chaudum:


MarkusH:
Let's merge #153 before issuing 0.12.2

chaudum:
> Let's merge #153 before issuing 0.12.2

#153 is a fix for an unreleased feature that is not part of version 0.12

PR #157, opened 2019-05-27
add region parameter to croud clusters delete command

m-kharbat:
## Checklist


croud/consumers/commands.py
```diff
@@ -102,7 +102,7 @@ def consumers_edit(args: Namespace) -> None:
 
     if config:
         body["config"] = config
-    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client = Client(env=args.env, region=args.region, output_fmt=args.output_fmt)
```
MarkusH: This is a bugfix to the 0.10 release. Thus requires a changelog entry. The other two changes are only in master.

PR #158, opened 2019-05-28
Replace setuptools_scm with manual version management

chaudum:

Croud used to use [setuptools_scm](https://pypi.org/project/setuptools-scm/)
for handling the Python package version instead of declaring it manually
in the `setup.py` file.

This commit removes the setup requirement and switches to manually
declared versions. The reason for that is to use the same tooling for
tagging and releasing this package, that we use also for other packages.
This change therefore also adds a script to create a tag automatically,
which is located in `dev/tools/create_tag.sh`.

## Checklist

### DIFF
--- a/DEVELOP.rst
+++ b/DEVELOP.rst
@@ -25,8 +25,8 @@ Run it::
 Git
 ---
 
-Alternatively, you can clone this repository, install it into a virtualenv and add the executable to your PATH
-environment variable::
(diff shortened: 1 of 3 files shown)


mfussenegger:
>  The reason for that is to use the same tooling for
tagging and releasing this package, that we use also for other packages.

Just curious - why not the other way around - change the others?

PR #165, opened 2019-06-03
print an empty table when the query results in an empty list

m-kharbat:
## Checklist

### DIFF
--- a/croud/printer.py
+++ b/croud/printer.py
@@ -93,11 +93,15 @@ class TableFormatPrinter(FormatPrinter):
         # |-----+-----|
         # | bar |   2 |
         # +-----+-----+
-        headers = list(map(str, rows[0].keys()))
+
+        headers = list(map(str, rows[0].keys())) if len(rows) else self.keys
+        if headers is None:
+            return ""
         values = [
             [self._transform_record_value(row[header]) for header in headers]
             for row in rows
         ]
+
         return tabulate(values, headers=headers, tablefmt="psql", missingval="NULL")
 
     def _filter_record(self, data: dict, keys: List[str]) -> JsonDict:
--- a/croud/rest.py
+++ b/croud/rest.py
@@ -105,7 +105,7 @@ class Client:
                 print_format(self._error, "json")
             return
 
-        if self._data is None or len(self._data) == 0:
+        if self._data is None:
             message = success_message or "Success."
             print_success(message)
             return
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Fixed an issue that caused empty query results to print "Success" to
+  the console instead of an empty table.
+
 0.13.1 - 2019/05/29
 ===================
 
--- a/tests/unit_tests/test_printer.py
+++ b/tests/unit_tests/test_printer.py
@@ -54,6 +54,17 @@ def test_json_format(rows, expected):
     (
         (None, ["what", "ever"], ""),
         ({}, ["what", "ever"], ""),
+        ([], [], ""),
+        (
+            [],
+            ["what", "ever"],
+            (
+                "+--------+--------+\n"
+                "| what   | ever   |\n"
+                "|--------+--------|\n"
+                "+--------+--------+"
+            ),
+        ),
         (
             {"a": "foo", "b": 1, "c": True},
             None,


croud/rest.py
```diff
@@ -97,7 +97,12 @@ def send(
             data = body["data"] if "data" in body else body
             return data, None
 
-    def print(self, success_message: str = None, keys: List[str] = None):
+    def print(
+        self,
+        success_message: str = None,
+        keys: List[str] = None,
+        level: str = Level.SUCCESS.value,
```
MarkusH: Let's make `level` a `Level` type, not a string. You can then pass the enum directly and not the string itself:
```suggestion
        level: Level = Level.SUCCESS,
```

croud/rest.py
```diff
@@ -107,7 +112,10 @@ def print(self, success_message: str = None, keys: List[str] = None):
 
         if self._data is None or len(self._data) == 0:
             message = success_message or "Success."
-            print_success(message)
+            if level and level == Level.INFO.value:
```
MarkusH: Since `level` has a default value and is not optional, you don't need to check for it being defined.
```suggestion
            if level == Level.INFO:
```

tests/unit_tests/test_rest.py
```diff
@@ -112,6 +113,21 @@ def test_print_success(mock_print_format, mock_print_success, event_loop):
...
+@mock.patch("croud.rest.print_format")
+def test_print_info(mock_print_format, mock_print_info, event_loop):
+    client = Client(env="dev", region="bregenz.a1", output_fmt="json", loop=event_loop)
+
+    client._data = {"key": "value"}
+    client.print(level=Level.INFO.value)
+    mock_print_format.assert_called_once_with({"key": "value"}, "json", None)
+
+    client._data = None
+    client.print(success_message="Info message.", level=Level.INFO.value)
```
MarkusH:
```suggestion
    client.print(success_message="Info message.", level=Level.INFO)
```

croud/organizations/commands.py
```diff
@@ -44,4 +45,8 @@ def organizations_list(args: Namespace) -> None:
 
     client = Client(env=args.env, output_fmt=args.output_fmt)
     client.send(RequestMethod.GET, "/api/v2/organizations/")
-    client.print(keys=["id", "name", "plan_type"])
+    client.print(
+        success_message="You do not belong to any organization",
+        keys=["id", "name", "plan_type"],
+        level=Level.INFO.value,
```
MarkusH:
```suggestion
        level=Level.INFO,
```

tests/unit_tests/test_rest.py
```diff
@@ -112,6 +113,21 @@ def test_print_success(mock_print_format, mock_print_success, event_loop):
     )
 
 
+@mock.patch("croud.rest.print_info")
+@mock.patch("croud.rest.print_format")
+def test_print_info(mock_print_format, mock_print_info, event_loop):
+    client = Client(env="dev", region="bregenz.a1", output_fmt="json", loop=event_loop)
+
+    client._data = {"key": "value"}
+    client.print(level=Level.INFO.value)
```
MarkusH:
```suggestion
    client.print(level=Level.INFO)
```

croud/rest.py
lukasbals: What happens if the level is `error`?

croud/rest.py
```diff
@@ -97,7 +97,12 @@ def send(
             data = body["data"] if "data" in body else body
             return data, None
 
-    def print(self, success_message: str = None, keys: List[str] = None):
+    def print(
+        self,
+        success_message: str = None,
```
lukasbals: maybe this could only be `message` ... because if the level is info it's no success-message

croud/rest.py
```diff
@@ -107,7 +112,10 @@ def print(self, success_message: str = None, keys: List[str] = None):
 
         if self._data is None or len(self._data) == 0:
             message = success_message or "Success."
-            print_success(message)
+            if level == Level.INFO:
```
lukasbals: What happens if the level is `error`?

croud/printer.py
```diff
@@ -93,11 +93,25 @@ def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
...
-        ]
+        if len(rows):
+            headers = list(map(str, rows[0].keys()))
+            values = [
+                [self._transform_record_value(row[header]) for header in headers]
+                for row in rows
+            ]
+
+        else:
+            # when there rows are empty we want to return en empty table, eg:
```
MarkusH:
```suggestion
            # when there are no rows but there is a list of keys we want to return an empty table:
```

croud/printer.py
```diff
@@ -93,11 +93,25 @@ def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
         # |-----+-----|
         # | bar |   2 |
         # +-----+-----+
-        headers = list(map(str, rows[0].keys()))
-        values = [
-            [self._transform_record_value(row[header]) for header in headers]
-            for row in rows
-        ]
+        if len(rows):
```
chaudum: IMO this could be even simpler:

```diff
diff --git a/croud/printer.py b/croud/printer.py
index d444d5e..d636cb0 100644
--- a/croud/printer.py
+++ b/croud/printer.py
@@ -93,7 +93,9 @@ class TableFormatPrinter(FormatPrinter):
         # |-----+-----|
         # | bar |   2 |
         # +-----+-----+
-        headers = list(map(str, rows[0].keys()))
+        headers = list(map(str, rows[0].keys())) if len(rows) else self.keys
+        if headers is None:
+            return ""
         values = [
             [self._transform_record_value(row[header]) for header in headers]
             for row in rows
```

PR #172, opened 2019-06-05
Remove current_context from configuration object

chaudum:

The Configuration class is responsible for reading from and writing to
the configuration file.
However it was also misused for "temporarily" storing the current
context so it is available across method calls but not persisted to
disk.
This PR removes the current_context property from the class.


chaudum:
retest this please

PR #179, opened 2019-06-06
standardize docs system

mechanomi:
this PR does the following:

- introduce a `Makefile` that provides a standard way to work with the docs. after this lands, I intend to port this same system over to the rest of our documentation projects

  my higher-level goal here is to standardize our docs system and have that live alongside whatever build system is being used for the rest of the project. i.e., this will exist alongside buildout for projects that use buildout. and it will be subsumed into Gradle for CrateDB (replacing the existing shell script we have that does this)

- hooks the doc tests into our Travis CI config

- improves the badges we're using in the README

- improves the requirements.txt we're using for the docs in a way that can also be mirrored across to other projects


mechanomi:
NOTE: I am having some issues configuring Travis CI to run the docs as a separate job. if I can't fix it tonight I will pick it back up tomorrow

Makefile
```diff
@@ -0,0 +1,63 @@
+# Documentation Utils
```
chaudum: Since this Makefile is docs specific, what about moving it into the docs/ folder?

README.rst
```diff
@@ -63,3 +43,23 @@ Help
...
+
+.. |version| image:: https://badge.fury.io/py/croud.svg
+    :target: http://badge.fury.io/py/croud
+    :alt: Version
+
+.. |docs| image:: https://readthedocs.org/projects/croud/badge/?version=latest
+    :alt: Documentation
+    :target: https://crate.io/docs/cloud/cli/en/latest/
+
+.. |coveralls| image:: https://coveralls.io/repos/github/crate/croud/badge.svg?branch=master
```
MarkusH: Why do we need `coveralls` when we have `codecov`? The two services provide pretty much the same feature set.

Makefile
mechanomi: @chaudum that's a great idea, thanks!

README.rst
mechanomi: @MarkusH I'm just re-formatting what's in the existing README. I don't want to make a change like that on this PR. (it's not my call, anyway)

are you happy with this PR as is?

if you (or someone else) removes Coveralls on `master` I will rebase my PR

mechanomi:
note to self: I want to move the Vale and doc8 components out of the Makefile in this repository and into a new Makefike in the crate-style-guide repository so that the only thing *this* project needs to specify is which version of the crate-style-guide is being used

docs/Makefile
```diff
@@ -0,0 +1,115 @@
...
+	@echo
+	@printf '\033[34m  test  \033[00m Test the documentation\n'
+
+$(SPHINXBUILD) $(SPHINXAUTOBUILD) $(DOC8):
+	'$(PYTHON)' -m venv '$(ENVDIR)'
+	'$(PIP)' install --upgrade $(PIPBUILDPKGS)
+	'$(PIP)' install -r '$(SRCDIR)/requirements.txt'
+	# Vale doesn't provide an option to skip directories, so we must delete
+	# third-party RST files from the Python venv
+	find '$(ENVDIR)' -name '*.rst' -delete
```
chaudum: Is this save?

mechanomi:
@chaudum ugh. the `docs/.test` submodule is a mistake. I was developing my approach to getting a tag of https://github.com/crate/crate-docs-style and at one point, I was storing it at this location. I will remove this from the PR

docs/Makefile
mechanomi: yup. `$(ENVDIR)` is an absolute path so it is guaranteed that all this will do is remove RST files (documentation) from the Python venv

PR #181, opened 2019-06-07
improved printing of error messages

m-kharbat:

the `errors` part of an error message was not printed, when the error contained a `message`.
this caused the schema validation error to be incomprehensible. 

## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Improved printing of error messages by including the error details returned
+  by the server.
+
 0.14.0 - 2019/06/06
 ===================
 
--- a/croud/rest.py
+++ b/croud/rest.py
@@ -102,6 +102,8 @@ class Client:
         if self._error:
             if "message" in self._error:
                 print_error(self._error["message"])
+                if "errors" in self._error:
+                    print_format(self._error["errors"], "json")
             else:
                 print_format(self._error, "json")
             return
--- a/tests/unit_tests/test_rest.py
+++ b/tests/unit_tests/test_rest.py
@@ -112,14 +112,15 @@ def test_print_success(mock_print_format, mock_print_success, event_loop):
     )
 
 
-@mock.patch("croud.rest.print_error")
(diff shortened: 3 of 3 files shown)


croud/rest.py
```diff
@@ -102,6 +102,8 @@ def print(self, success_message: str = None, keys: List[str] = None):
         if self._error:
             if "message" in self._error:
                 print_error(self._error["message"])
+                if "errors" in self._error:
```
MarkusH: I don't understand why we only print the errors when we also have a message? Why not always print the errors when they're present?

tests/unit_tests/test_rest.py
```diff
@@ -112,14 +112,16 @@ def test_print_success(mock_print_format, mock_print_success, event_loop):
     )
 
 
+@mock.patch("croud.rest.print_format")
 @mock.patch("croud.rest.print_error")
-def test_print_error(mock_print_error, event_loop):
+def test_print_error(mock_print_error, mock_print_format, event_loop):
```
MarkusH: I think we should switch to using the [`capsys` fixture](https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function) over mocking the print functions. That way we get a clear picture of what the response looks like.

croud/rest.py
m-kharbat: in line 108 the full error is printed. 
The problem is that when we had the ``message`` we printed only the message. 
When we don't have the `` message`` the full error response is printed.

tests/unit_tests/test_rest.py
m-kharbat: due to the colours, I find it more readable to mock the print functions.

MarkusH: I think I agree. Good point. Let's check for `Bad request` _and_ `"key": "Error on \'key\'"`, though.

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Improved printing of error messages.
```
```suggestion
- Improved printing of error messages by including the error details returned by the server.
```

PR #185, opened 2019-06-12
Updated projects users add/remove commands from GraphQL to REST

lukasbals:

Updated the `projects users add/remove` commands from GraphQL to REST.

This PR depends on changes in the backend. https://github.com/crate/cloud-app/pull/835

## Checklist


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Updated ``projects users add/remove`` commands from graphql to rest
```
```suggestion
- Updated ``projects users add/remove`` commands from GraphQL to REST.
```

croud/projects/users/commands.py
```diff
@@ -19,48 +19,33 @@
...
-    query = Query(_query, args)
-    query.execute()
-    print_query(query, "addUserToProject", "Successfully added user to project.")
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.POST,
+        f"/api/v2/projects/{args.project_id}/users",
+        body={"user": args.user},
+    )
+    client.print(keys=["user_id", "role_fqn", "resource_id"])
```
MarkusH: I think that should be `organization_id` and not `organization_id`, shouldn't it: https://github.com/crate/cloud-app/blob/3b7b127a7383cd2e6cb6a0037d2c11cdf5c8477d/app/account/organizations/schemas.py#L24 ?

croud/projects/users/commands.py
```diff
@@ -19,48 +19,33 @@
...
-    query.execute()
-    print_query(
-        query, "removeUserFromProject", "Successfully removed user from project."
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.DELETE,
+        f"/api/v2/projects/{args.project_id}/users",
+        body={"user": args.user},
     )
+    client.print(keys=["user_id", "role_fqn", "resource_id"])
```
MarkusH: As above, this should be `organization_id`, not `resource_id`.

croud/projects/users/commands.py
```diff
@@ -19,48 +19,33 @@
...
-    }}
-    """
-
-    query = Query(_query, args)
-    query.execute()
-    print_query(query, "addUserToProject", "Successfully added user to project.")
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.POST,
+        f"/api/v2/projects/{args.project_id}/users",
```
MarkusH:
```suggestion
        f"/api/v2/projects/{args.project_id}/users/",
```

croud/projects/users/commands.py
```diff
@@ -19,48 +19,33 @@
...
     """
 
-    query = Query(_query, args)
-    query.execute()
-    print_query(
-        query, "removeUserFromProject", "Successfully removed user from project."
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.DELETE,
+        f"/api/v2/projects/{args.project_id}/users",
```
MarkusH:
```suggestion
        f"/api/v2/projects/{args.project_id}/users/{args.user}/",
```

croud/projects/users/commands.py
```diff
@@ -19,48 +19,33 @@
...
+        body={"user": args.user},
```
MarkusH:
```suggestion

```

croud/projects/users/commands.py
lukasbals: It depends on Meriam's implementation, but I think `project_id` would be best 🤔

croud/projects/users/commands.py
```diff
@@ -19,48 +19,31 @@
...
 
-    query = Query(_query, args)
-    query.execute()
-    print_query(
-        query, "removeUserFromProject", "Successfully removed user from project."
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.DELETE, f"/api/v2/projects/{args.project_id}/users/{args.user}/"
     )
+    client.print(keys=["user_id", "role_fqn", "project_id"])
```
lukasbals: `success_message` instead of keys.

croud/projects/users/commands.py
```diff
@@ -19,48 +19,31 @@
...
+    Removes a user from a project.
     """
 
-    query = Query(_query, args)
-    query.execute()
-    print_query(
-        query, "removeUserFromProject", "Successfully removed user from project."
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.DELETE, f"/api/v2/projects/{args.project_id}/users/{args.user}/"
```
MarkusH: Let's use proper URL escaping as in https://github.com/crate/croud/pull/186/files#diff-fc96a85f2127221b6aa3b3a312e082a9R81 (same for above).

chaudum: `yarl` ftw!

lukasbals: Isn't it enough that yarl is used here?
https://github.com/crate/croud/blob/1c7609bb7b3bab4080dcebdbee661c5cb0a21c0b/croud/session.py#L81

croud/projects/users/commands.py
```diff
@@ -19,48 +19,31 @@
...
-    query = Query(_query, args)
-    query.execute()
-    print_query(query, "addUserToProject", "Successfully added user to project.")
+    client = Client(env=args.env, output_fmt=args.output_fmt)
+    client.send(
+        RequestMethod.POST,
+        f"/api/v2/projects/{args.project_id}/users/",
+        body={"user": args.user},
+    )
+    client.print(keys=["user", "role_fqn"])
```
MarkusH: Contrary to the organization's user adding, this doesn't return the project ID. Let's change that.

lukasbals: So you want to update the backend to return the `project_id` in addition?

MarkusH: Err, sorry. I meant "print", not "return".
```suggestion
    client.print(keys=["project_id", "user", "role_fqn"])
```

docs/commands/projects.rst
```diff
@@ -84,7 +84,11 @@ Example
    sh$ croud projects users add \
        --project-id 035f1161-402e-44b4-9073-0749586091e0 \
        --user 6ac0f500-f9f8-4c12-82e2-3ad6192525d4
-   ==> Success: Successfully added user to project.
+   +----------------+--------------------------------------+
```
chaudum: Don't we also print the project_id?
https://github.com/crate/croud/pull/185/files#diff-f22b31e19e62062d6c0d9e563e98a912R37

CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Updated ``projects users add/remove`` commands from GraphQL to REST.
```
chaudum: This is only a user-facing change in case the result changes. Does it? If so, please state that the result of the `projects users add/remove` commands changed.

docs/commands/projects.rst
lukasbals: I'm working on that already. It's not returned from the backend ATM. There is a PR for the backend open.

CHANGES.rst
```diff
@@ -5,6 +5,10 @@ Changes for croud
 Unreleased
 ==========
 
+- Updated ``projects users add/remove`` commands from GraphQL to REST. The result of the
```
chaudum:
```suggestion
- The response of the ``projects users add`` command changed from displaying only a success message to
```

croud/projects/users/commands.py
```diff
@@ -19,48 +19,32 @@
...
-    mutation {{
-        addUserToProject(input: {{
-            projectId: "{args.project_id}",
-            user: "{args.user}"
-        }}) {{
-            success
-        }}
-    }}
-    """
+    client = Client(env=args.env, output_fmt=args.output_fmt)
```
MarkusH: I think we need to pass `args.region` as well, don't we?

croud/projects/users/commands.py
```diff
@@ -19,48 +19,32 @@
...
-        }}
-    }}
+    Removes a user from a project.
     """
 
-    query = Query(_query, args)
-    query.execute()
-    print_query(
-        query, "removeUserFromProject", "Successfully removed user from project."
+    client = Client(env=args.env, output_fmt=args.output_fmt)
```
MarkusH: I think we need to pass `args.region` as well, don't we?

croud/projects/users/commands.py
Taliik: not sure if we need to, i think we take the region that the user is currently in. if he wants to create/delete something in a different region he could switch first

Taliik: this means all our commands need to be updated to pass the region here. they are currently all wrong :(

chaudum: Let's change it here and make another PR which fixes all other occurrences.

PR #190, opened 2019-07-01
Use token from selected environment for authentication

MarkusH:

Before, the token from the `current_context` was always used for authentication purposes.

## Checklist


croud/config.py
```diff
@@ -106,8 +106,9 @@ def set_context(env: str) -> None:
         write_config(config)
 
     @staticmethod
-    def get_token() -> str:
-        return get_auth_context().get("token", "")
+    def get_token(env: str) -> str:
+        config = load_config()
+        return config["auth"]["contexts"][env].get("token", "")
```
chaudum: This could cause a KeyError, however, not in the current usage within croud.
I leaf it up to you if you want to raise a more meaningful exception in that case.

MarkusH: I don't think it can raise a KeyError. `get_token` calls `load_config` which performs a config validation. And a valid config has all the keys listed here.

chaudum: However, `env` is variable and could be anything (when looking at the function without the context of croud).

chaudum: As I said, when using croud `env` can only be a valid (existing) key (prod, dev, local).

PR #200, opened 2019-07-05
Mh/rest users roles add remove

MarkusH:

* Rename command functions for user adding/removing to projects to plural
  This follows along with the naming convention we have for other commands as well.
* Updated `users roles add/remove` to use the REST API
  Under the hood, neither command is useful anymore. There's already a clearer and more explicit interface in `croud organizations/projects users add/remove`.
  There's a tiny side-effect though, with the switch from GraphQL to the REST API: the REST API doesn't take the role into account when removing a user from an organization or project. This is, because a user can only ever have one role per resource.
* Use email address in docs when adding/removing users to/from orgs/projects

## Checklist


tests/unit_tests/test_commands.py
```diff
@@ -562,71 +562,93 @@ def test_remove(self, mock_send, mock_load_config):
...
-        user_id = gen_uuid()
-        role_fqn = "org_admin"
+    @pytest.mark.parametrize(
+        "role_fqn,user",
+        [
+            ("org_admin", gen_uuid()),
+            ("org_member", "foo@crate-dev.cloud"),
+            ("project_admin", "bar@crate-dev.cloud"),
+            ("project_member", gen_uuid()),
+            ("invalid", gen_uuid()),
```
chaudum: Please make a separate test case for an invalid role.

PR #212, opened 2019-07-19
Remove mypy.ini and tox.ini in favour of setup.cfg

chaudum:


lukasbals:
retest this please

chaudum:
> I've never seen the tox config in `setup.cfg` which makes this setup a bit weird. But I also didn't know it was possible in the first place.

I guess we can keep the tox.ini for this purpose. I haven't seen it anywhere else either.

MarkusH:
> > I've never seen the tox config in `setup.cfg` which makes this setup a bit weird. But I also didn't know it was possible in the first place.
> 
> I guess we can keep the tox.ini for this purpose. I haven't seen it anywhere else either.

Sounds good. Having a `tox.ini` and moving the `mypy.ini` into `setup.cfg`.

PR #213, opened 2019-07-23
Replace function-based argument adding by declarative style instances

MarkusH:

* I ran `tox -e py37` after each commit successfully, so each commit passes individually
* I think we should squash all commits together.

## Checklist


tests/unit_tests/test_parser.py
```diff
@@ -115,6 +106,10 @@ def argument_provider(req_group, opt_group):
                 parser.parse_args(["cmd"])
             assert ex_info.value.code == 2
             output = stdout.call_args[0][0]
+
+        print("\n\n\n")
```
chaudum: Are these print statements handy? If so, I would recommend them into the `assert_ellipsis_match` method.

MarkusH: Oops, nope, they're left overs from debugging

PR #216, opened 2019-07-26
Added the --sudo argument

Taliik:
## Checklist

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Added the optional ``--sudo`` argument to be able to specify when a command should
+  be run as ``superuser``.
+
 - The ``-org-id`` parameter is now optional for non-superusers. Upon login, a
   user's organization will be retrieved and stored in the configuration file.
   Whenever the ``--org-id`` parameter is needed, a fallback to the default
--- a/croud/rest.py
+++ b/croud/rest.py
@@ -18,6 +18,7 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import asyncio
+import uuid
 from argparse import Namespace
 from functools import partial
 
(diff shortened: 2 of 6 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,9 @@ Changes for croud
 Unreleased
 ==========
 
+- Added the optional ``--sudo`` argument to be able to specify if a command should
```
```suggestion
- Added the optional ``--sudo`` argument to be able to specify when a command
  should
```

tests/unit_tests/test_rest.py
```diff
@@ -47,23 +47,29 @@ def test_send_success_sets_data_with_key(self, mock_cloud_url, mock_token):
         with mock.patch.object(HttpSession, "_get_conn", return_value=self._get_conn()):
             data = {"key": "value"}
 
-            client = Client(env="dev", region="bregenz.a1", output_fmt="json")
+            client = Client(
+                env="dev", region="bregenz.a1", output_fmt="json", sudo=False
```
MarkusH: No need to pass `sudo=False` here all the time.

croud/rest.py
```diff
@@ -34,20 +34,24 @@ class Client:
...
     _token: str
     _region: str
+    _sudo: bool
     _output_fmt: str
     _data: Optional[Union[List[dict], dict]] = None
     _error: Optional[dict] = None
 
-    def __init__(self, env: str, region: str, output_fmt: str, loop=None):
+    def __init__(
+        self, env: str, region: str, output_fmt: str, sudo: bool = False, loop=None
```
MarkusH: Let's add `sudo` at the end to keep the argument order. Or make `loop` and `sudo` keywork only arguments:
```python
def __init__(
    self, env: str, region: str, output_fmt: str, loop=None, sudo: bool = False
):
```
```python
def __init__(
    self, env: str, region: str, output_fmt: str, *, loop=None, sudo: bool = False
):
```

Taliik:
> I think the former was correct. Now you're sending `X-Auth-Sudo: 0` which is an invalid header value.

So i should just check that the header is not there at all when sudo=False i suppose

PR #217, opened 2019-07-29
Add support to list users per organization, project and update the system-wide listing

MarkusH:

- Add `croud organizations users list`
- Add `croud projects users list`
- Update `croud users list` to show user roles and allow filtering for users w/o any roles.

## Checklist


croud/cmd.py
```diff
@@ -39,6 +39,17 @@ def project_name_arg(req_args: _ArgumentGroup, opt_args: _ArgumentGroup) -> None
     )
 
 
+def no_roles_arg(req_args: _ArgumentGroup, opt_args: _ArgumentGroup) -> None:
+    opt_args.add_argument(
+        "--no-roles",
+        nargs="?",
```
chaudum: IMO an argument with `action="store_true"` is a better fit for this type of argument.

croud/__main__.py
```diff
@@ -410,7 +437,12 @@
             "list": {
                 "help": "List all users within the specified organization.",
```
chaudum: I think we should update the help text of the command.

croud/organizations/users/commands.py
```diff
@@ -39,6 +39,20 @@ def org_users_add(args: Namespace):
...
+    """
+    Lists organization users
+    """
+
+    client = Client.from_args(args)
+    client.send(RequestMethod.GET, f"/api/v2/organizations/{args.org_id}/users/")
+    if client._data and client._output_fmt in {"table"}:
+        for user in client._data:
+            user["role_fqn"] = user["organization_roles"][0]["role_fqn"]
+    client.print(keys=["uid", "email", "username", "role_fqn"])
```
chaudum: What do you think about changing the signature of `client.print()` to also allow a list of "transformer" functions for the specified keys:

```suggestion
    client.print(
        keys=["uid", "email", "username", "role_fqn"],
        transform=[identity, identity, identity, transform_org_role],
    )
    
def identity(key, item):
    return item[key]
    
def transform_org_role(key, item):
    return item["organization_roles"][0][key]
```

chaudum: Same applies for the other transformations.

MarkusH: I like the idea.

MarkusH:
I added the transforms and updated the docs.

MarkusH:
> 1st and 2nd commit need to be squashed because the 1st commit requires code of the 2nd.

Nope, that's an illusion due to GitHub's temporal sorting of commits, and not hierarchical sorting. Running `tox -e py37` after each commit works perfectly fine.

PR #220, opened 2019-07-30
Fixed the error when using the format argument

Taliik:
## Checklist

### DIFF
--- a/croud/rest.py
+++ b/croud/rest.py
@@ -44,7 +44,7 @@ class Client:
 
     @staticmethod
     def from_args(args: Namespace) -> "Client":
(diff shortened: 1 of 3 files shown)


croud/rest.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, env: str, region: str, loop=None, sudo: bool = False):
 
     @staticmethod
     def from_args(args: Namespace) -> "Client":
-        return Client(args.env, args.region, args.output_fmt, args.sudo)
+        return Client(args.env, args.region, args.sudo)
```
MarkusH:
```suggestion
        return Client(args.env, args.region, sudo=args.sudo)
```

PR #221, opened 2019-07-30
Upgrade docs utils to 0.1.11

mechanomi:


DEVELOP.rst
```diff
@@ -79,37 +84,80 @@ See the instructions for `Generating distribution archives`_ for more details.
...
+    $ make help
 
-    $ pip install -Ur requirements-docs.txt
+    Crate Documentation Build System
 
-Then build the documentation by running::
+    Run `make <TARGET>`, where <TARGET> is one of:
 
-    $ bin/sphinx
+      dev     Run a Sphinx development server that builds and lints the
```
Taliik: Running `make dev` (also `make check`) results in library description being linted e.g.
`/env/lib/python3.7/site-packages/schema-0.6.8.dist-info/DESCRIPTION.rst`, which results in an error. Is it intended to run lint over library entries?

Taliik:
Also, can you please add the `.utils` folder which is created when executing e.g. `make help` to the git ignore :)

DEVELOP.rst
mechanomi: nope. fixed in the latest version of the docs utils. thx!

mechanomi:
@Taliik pushed a rebased fixup. can you recheck? thanks!

mechanomi:
@Taliik thanks! the `.build` directory is no longer generated. if you have a copy locally, you can get rid of it safely

PR #222, opened 2019-07-30
Fixed a bug using the croud config commands

Taliik:
## Checklist

### DIFF
--- a/croud/parser.py
+++ b/croud/parser.py
@@ -111,6 +111,9 @@ def add_default_args(parser):
         choices={"table", "json", "yaml"},
         help="Change the formatting of the output",
     )
+
+
+def add_default_sudo_arg(parser):
     parser._group_optional.add_argument(
         "--sudo",
         required=False,
@@ -142,6 +145,8 @@ def add_subparser(parser, tree, name="__root__"):
             add_subparser(sub, _tree, sub.prog)
     else:
         add_default_args(parser)
+        if not tree.get("omit_sudo", False):
+            add_default_sudo_arg(parser)
         resolver = tree["resolver"]
         signature = inspect.signature(resolver)
         if "parser" in signature.parameters:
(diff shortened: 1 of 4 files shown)


croud/parser.py
```diff
@@ -143,6 +146,10 @@ def add_subparser(parser, tree, name="__root__"):
     else:
         add_default_args(parser)
         resolver = tree["resolver"]
+        # This check is needed so that the sudo argument is not present in the
+        # config_set/config_get command, where it should not be available.
+        if resolver.__name__ not in ["config_set", "config_get"]:
```
MarkusH: Checking for the function name here seems strange. How about adding an optional `has_sudo` key to the command dictionary that defaults to sudo in `add_subparser` and is set to `False` for the config commands?

```python
command_tree = {
    "config": {
        "help": "Manage croud default configuration values.",
        "commands": {
            "get": {
                "help": "Get default configuration values.",
                "resolver": config_get,
                "noop_arg": {"choices": ["env", "region", "output-fmt"]},
                "has_sudo": False,
            },
            "set": {
                "help": "Set default configuration values.",
                "resolver": config_set,
                "has_sudo": False,
            },
        },
    },
}

def add_subparser(parser, tree, name="__root__"):
    ...
    if "commands" in tree:
        ...
    else:
        add_default_args(parser)
        if tree.get("has_sudo", True):
            add_default_sudo_arg(parser)
        ...
```

Taliik: Seems like a nice idea aswell

chaudum: :+1: for the idea
Personally I would "reverse" the optional key in the declaration tree, e.g. name it `omit_sudo`.
The reason for that is, that I prefer treating optional, not present keys as falsy values.
But that is only my take on it. Feel free to ignore it :)

PR #227, opened 2019-08-02
F/org edit command

Taliik:
- The ``plan-type`` argument has been made optional and is only available for superusers.

- Added new ``croud organizations edit`` command to edit an existing organization.

## Checklist


croud/organizations/commands.py
```diff
@@ -32,16 +32,46 @@ def organizations_create(args: Namespace) -> None:
...
+
+
+@org_id_config_fallback
+def organizations_edit(args: Namespace) -> None:
+    """
+    Edit an organization
+    """
+
+    client = Client.from_args(args)
+    if args.plan_type and args.name:
```
chaudum:
```suggestion
    body = {}
    if args.plan_type:
        body["plan_type"] = args.plan_type
    if args.name:
        body["name"] = args.name
    if not body:
        print_error("No input arguments provided.")
```

croud/__main__.py
```diff
@@ -468,6 +469,25 @@
...
+                "help": "Edit the specified organization.",
+                "extra_args": [
+                    Argument(
+                        "--name", type=str, required=False,
+                        help="The new organization name to use.",
+                    ),
+                    Argument(
+                        "--plan-type", type=int, required=False,
+                        choices=[1, 2, 3, 4, 5, 6],
+                        help="The new support plan to use for the organization.",
```
chaudum: Should we add the limitations to the help text?

```suggestion
                        help="The new support plan to use for the organization. Argument is for superusers only.",
```

/cc @lukas-bernert

lukas-bernert: I'd be in favor of doing it. However, we already have commands that are superuser only (e.g. croud monitoring..). If we want to implement it here we should be consistent and also think about adding it to other superuser-only commands. Either now or in the future.

chaudum: I think there is a difference between superuser commands, and superuser arguments. Afaik, `--plan-type` is the only argument that is only relevant for superusers. Everywhere else, the whole command is superuser only.
I don't mind, not adding the limitation to the help text for now.

lukas-bernert: True. I'd add it for now. It helps for a better UX IMO.

PR #228, opened 2019-08-02
Fix bug in organization delete

Taliik:
## Checklist

### DIFF
--- a/tests/unit_tests/test_commands.py
+++ b/tests/unit_tests/test_commands.py
@@ -41,6 +41,21 @@ def gen_uuid() -> str:
     return str(uuid.uuid4())
 
 
+config_org_id = gen_uuid()
+FALLBACK_ORG_ID_CONFIG: dict = {
+    "auth": {
+        "current_context": "prod",
+        "contexts": {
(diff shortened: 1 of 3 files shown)


tests/unit_tests/test_commands.py
```diff
@@ -383,6 +400,52 @@ def test_delete(self, mock_send, mock_load_config, capsys):
...
+    ):
+        argv = ["croud", "organizations", "delete"]
+        with mock.patch("builtins.input", side_effect=["Y"]) as mock_input:
+            self.assertRest(
+                mock_send,
+                argv,
+                RequestMethod.DELETE,
+                f"/api/v2/organizations/{mock_load_config.return_value['auth']['contexts']['local']['organization_id']}/",  # noqa
+            )
+            mock_input.assert_called_once_with(
```
chaudum: Don't think we need to assert this, because it's already done in the `test_delete ` method.

tests/unit_tests/test_commands.py
```diff
@@ -383,6 +400,52 @@ def test_delete(self, mock_send, mock_load_config, capsys):
...
```
chaudum: Same here.

PR #235, opened 2019-08-08
[WiP] Remove asyncio

Taliik:
First try at removing asyncio in favor of a simple HTTPServer and requests.

## Checklist


croud/transport.py
```diff
@@ -0,0 +1,69 @@
+import requests
+import uuid
+from argparse import Namespace
+from functools import partial
+
+from aiohttp import ContentTypeError  # type: ignore
```
MarkusH: Can we not depend on aiohttp?

croud/transport.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def _decode_response(self, resp: requests.Response):
+        if resp.status_code == 204:
+            # response is empty
+            return None, None
+
+        try:
+            body = resp.json()
+        # API always returns JSON, unless there's an unhandled server error
+        except ContentTypeError:
```
MarkusH: I think that's a different error that you'll get than one from aiohttp.

croud/server.py
```diff
@@ -17,51 +17,66 @@
...
-        await self.runner.setup()
-        site = web.TCPSite(self.runner, "localhost", Server.PORT)
-        await site.start()
 
-    async def stop(self) -> None:
-        """Shutdown the server gracefully"""
-        await self.runner.cleanup()
+def run_server(on_token) -> Thread:
+    server = SetTokenHTTPServer(on_token, (HOST, PORT), SetTokenHandler)
+    server_thread = Thread(target=server.serve_forever, daemon=True)
```
MarkusH: Why do we need an extra thread? The server already runs in one, AIUI.

PR #256, opened 2019-12-19
Remove asyncio and upgrade dependencies

MarkusH:
## Checklist


croud/server.py
```diff
@@ -17,51 +17,78 @@
...
+        query_string = parse.urlparse(self.path).query
+        query = parse.parse_qs(query_string)
+        if "token" not in query:
+            code = 400
+            msg = SetTokenHandler.MISSING_TOKEN_MSG
+        elif len(query["token"]) != 1:
+            code = 400
+            msg = SetTokenHandler.DUPLICATE_TOKEN_MSG
+        else:
+            token = query["token"][0]
```
chaudum: Should we account for the edge case that the token parameter is present but the token is an empty string?

tests/conftest.py
```diff
@@ -1,11 +1,50 @@
...
-from tests.util.fake_server import FakeCrateDBCloud, FakeResolver
+from .util.fake_cloud import FakeCrateDBCloud
+
+
+@pytest.fixture(scope="session")
+def fake_cratedb_cloud():
+    cloud = FakeCrateDBCloud()
+    cloud.start()
+    yield cloud
+    cloud._server.shutdown()
```
chaudum: Instead of accessing a "protected" property, I would provide a method to stop the FakeCrateDBCloud, or it could even be a context manager type:

```suggestion
    with FakeCrateDBCloud() as cloud:
        yield cloud
```

tests/conftest.py
```diff
@@ -1,11 +1,50 @@
...
+
+    with mock.patch(
+        "croud.api.cloud_url",
+        return_value=f"https://127.0.0.1:{fake_cratedb_cloud.port}",
+    ):
+        with mock.patch("croud.api.get_verify_ssl", return_value=False):
+            with mock.patch(
+                "croud.api.Configuration.get_token", return_value="some-token"
+            ):
+                yield Client(env="local", region="bregenz.a1")
```
chaudum: As described above, I would use a dedicated `verify_ssl` keyword instead:

```suggestion
                yield Client(env="local", region="bregenz.a1", verify_ssl=False)
```

setup.py
```diff
@@ -61,5 +61,6 @@
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
```
chaudum: I think we should add the Python3.8 support to the changelog.

croud/api.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def request(
+        self,
+        method: RequestMethod,
+        endpoint: str,
+        *,
+        params: dict = None,
+        body: dict = None,
+    ):
+        kwargs: dict = {"allow_redirects": False}
```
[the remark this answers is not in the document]
MarkusH: No, otherwise mypy derives an implicit `Dict[str, bool]` which clashes with the next lines of like `kwargs["params"] = params` as `params` is `Dict[Any, Any]`. Using `dict` makes it a `Dict[Any, Any]`.

croud/server.py
MarkusH: https://github.com/crate/croud/issues/257

croud/api.py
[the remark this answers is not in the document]
chaudum: Right, ok. Sometimes mypy is more complicated than necessary :D

PR #260, opened 2020-02-06
Add note that cluster deployments are only available to superusers

MarkusH:

Do we want a link to the Azure offer in the note to give readers a way to continue?

## Checklist


matthijskrul:
For now I think this would be a good idea. I think once it is incorporated into the new Cloud docs (rather than Intercom), we should then (also) link to the equivalent of this: https://help.crate.io/en/articles/3603380-how-to-deploy-a-cluster-via-the-azure-marketplace

PR #264, opened 2020-03-19
Always use API region from config to make API calls

chaudum:
This commit changes the way how the `--region` argument is used for API
calls. Where previously this argument temporarily overwrote the `region`
setting from the configuration to invoke HTTP requests on the specified
region specific subdomain of `cratedb[-dev].cloud`, API requests are now
always invoked on the region specified in the configuration. However,
the HTTP request header `X-Region` is set to the value specified using
the `--region` argument, or if not present, to the value of the region
configuration setting.
This is however not the case for the `croud config set` command! This
command still sets the configuration value.

*This change requires a CrateDB Cloud version that implements the
functionality of the `X-Region` HTTP header.*


croud/api.py
```diff
@@ -47,21 +47,29 @@ def __init__(
         self, *, env: str, region: str, sudo: bool = False, _verify_ssl: bool = True
     ):
         """
-        :param bool _verify_ssl: A private variable that must only be used during tests!
+        :param str env:
+          CrateDB Cloud environment, either "dev" or "prod"
+        :param str region:
+          Region from which to get data from or to which to deploy (defines the
```
matthijskrul: I'd just write "Region to get data from" etc.

tests/test_api.py
```diff
@@ -70,22 +70,23 @@ def test_send_new_token_response(client: Client):
...
 
-# If ``sudo=True``, the X-Auth-Sudo header should be set with any value. This
-# test checks that the header is set.
-def test_send_sudo_header_set(client: Client):
-    client = Client(env="dev", region="bregenz.a1", sudo=True, _verify_ssl=False)
-    resp_data, errors = client.get("/test-x-sudo")
-    assert resp_data == {}
-    assert errors is None
+def test_send_sudo_header(client: Client):
+    for sudo in [True, False]:
```
MarkusH: Can we use test parametrization for this, please. Otherwise one can't that easily distinguish which test failed, the one for True or the one for False.

tests/test_api.py
```diff
@@ -70,22 +70,23 @@ def test_send_new_token_response(client: Client):
...
-    assert resp_data == {"message": "Header not set, as expected."}
-    assert errors is None
+@pytest.mark.parametrize(
+    "argument,header", [(None, "bregenz.a1"), ("westeurope.azure", "westeurope.azure")]
+)
+def test_send_region_header(argument, header, client: Client):
+    client = Client(env="dev", region=argument, _verify_ssl=False)
+    resp_data, errors = client.get("/client-headers")
+    assert "X-Region" in [key for key, value in resp_data]  # type: ignore
+    assert dict(resp_data)["X-Region"] == header  # type: ignore
```
MarkusH: When the `client_headers()` function below returns a jsonified dict instead of a list of 2-tuples, the assertion here and in `test_send_sudo_header` become simpler:
```python
return Response(json_data=dict(self.headers.items()))
```
```suggestion
    assert resp_data["X-Region"] == header
```
Using key access on the response data guarantees the item is in the response and that it has the given value.

Similarly, above, it'd be
```python
assert resp_data["X-Auth-Sudo"] == sudo
```

croud/api.py
```diff
@@ -47,21 +47,29 @@ def __init__(
...
+        api_region = Configuration.get_setting("region")
+        self.base_url = URL(construct_api_base_url(self.env, api_region))
         self._token = Configuration.get_token(self.env)
+
         self.session = requests.Session()
         if _verify_ssl is False:
             self.session.verify = False
         self.session.cookies["session"] = self._token
-        if self.sudo:
+        self.session.headers["X-Region"] = region or api_region
```
MarkusH: This will effectively always send the header, even when it's the same as the region in the domain. Is that deliberate?

chaudum: Yes it is :)

PR #265, opened 2020-03-19
Use Github Actions for running tests

chaudum:
Documentation is still linted on TravisCI

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,2 @@
-pytest-black==0.3.7
-pytest-flake8==1.0.4
-pytest-isort==0.3.1
-pytest-mypy==0.4.2
-pytest-random-order==1.0.4
 pytest>=5,<6
+pytest-random-order==1.0.4
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,69 @@
+---
+name: CI
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  lint:
+    name: Code style checks and static analysis with Python ${{ matrix.python-version }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version:
+        - 3.8
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install dependencies
+        run: |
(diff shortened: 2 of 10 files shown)


test-requirements.txt
```diff
@@ -1,6 +1,2 @@
-pytest-black==0.3.7
```
christianbader: Removing this stuff breaks my local setup. Don't really like it.

chaudum: In which way? These dependencies have been only installed into tox's virtualenv when running `tox`.

christianbader: Yes, sorry for the noise. It never worked for this project 😞

chaudum: Ah I think I see your point. I pushed a fixup with `mypy`, `black`, `flake8`, and `isort` dependencies in the `development` extras.

```
$ python -m pip install -Ue ".[develpment"
```

christianbader: Thank you!

christianbader:
> What's the reason for not moving the macOS tests and documentation tests to GitHub Actions? OSX is available for GHA: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources

With macOS tests you mean running the tests on a macOS host? Don't think that this is neccessary tbh.

chaudum:
> > What's the reason for not moving the macOS tests and documentation tests to GitHub Actions? OSX is available for GHA: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources
> 
> With macOS tests you mean running the tests on a macOS host? Don't think that this is neccessary tbh.

I would leave that up to the @crate/docs team because they need to maintain it.

MarkusH:
> > > What's the reason for not moving the macOS tests and documentation tests to GitHub Actions? OSX is available for GHA: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources
> > 
> > 
> > With macOS tests you mean running the tests on a macOS host? Don't think that this is neccessary tbh.
> 
> I would leave that up to the @crate/docs team because they need to maintain it.

We're already running tests on macOS, but currently on Travis.

For docs, yeah, leaving that to the docs team seems fine with me.

chaudum:
> We're already running tests on macOS, but currently on Travis.

No, only documentation is linted on macOS
https://travis-ci.org/github/crate/croud/builds/664757215

MarkusH:
> > We're already running tests on macOS, but currently on Travis.
> 
> No, only documentation is linted on macOS
> https://travis-ci.org/github/crate/croud/builds/664757215

:see_no_evil:

PR #273, opened 2020-04-07
Add Github action to build & upload to PyPi

chaudum:

The Github action job is triggered whenever a tag is pushed. It builds
the source distribution package and the wheel and uploads them to PyPi.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,26 @@
+---
+name: Release
+on: push
+jobs:
+  pypi:
+    name: Build and publish to PyPI
+    if: startsWith(github.event.ref, 'refs/tags')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install setuptools and wheel
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,26 @@
...
+        with:
+          python-version: 3.8
+      - name: Install setuptools and wheel
+        run: |
+          python -m pip install --upgrade pip setuptools wheel
+      - name: Build sdist and wheel
+        run: |
+          python setup.py sdist bdist_wheel
+      - name: Publish package to PyPI
+        uses: pypa/gh-action-pypi-publish@master
```
MarkusH, line 23: Can we pin this to a commit, please.

DEVELOP.rst
```diff
@@ -125,15 +125,16 @@ Release
     ./dev/tools/create_tag.sh
 
 
-Upload to `PyPI`_
------------------
+Upload to PyPI
+--------------
 
-See the instructions for `Generating distribution archives`_ for more details.
+Pushing a tag to Github will trigger a Github workflow that builds and upload
```
matthijskrul, line 131: uploads*

.github/workflows/release.yaml
chaudum: There is actually already a release version: https://github.com/pypa/gh-action-pypi-publish/releases/tag/v1.1.0

I am wondering why almost all documentation of Github actions always refer to `@master` - I guess to receive updates automatically. Not sure if this is really good practice as you pointed out.

PR #278, opened 2020-05-06
Ensure config directory exists before saving config

MarkusH:
## Checklist

### DIFF
--- a/croud/config/configuration.py
+++ b/croud/config/configuration.py
@@ -70,8 +70,8 @@ class Configuration:
     """
 
     def __init__(self, name: str, path: Optional[Path] = None):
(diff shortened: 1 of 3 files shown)


croud/config/configuration.py
```diff
@@ -136,6 +136,7 @@ def is_valid(self):
     def dump(self) -> None:
         # make sure the config is in memory before we open the file for writing
         data = self.config
+        self._config_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
```
chaudum, line 139: Does the `mode` have any effects on Windows?

MarkusH: No, doesn't make a difference.

PR #283, opened 2020-05-14
Warning that users commands are for superusers only

matthijskrul:
Adds warning to Croud docs section on user commands that these are for Cloud staff (superusers) only and are only listed because they are currently present in the argparse configuration for --help. Corresponds to [this old story/impediment](https://trello.com/c/2uUClYL5/251-fix-customer-facing-docs-for-croud-users-command). 

## Checklist

### DIFF
--- a/docs/commands/users.rst
+++ b/docs/commands/users.rst
@@ -15,6 +15,13 @@ The ``users`` command allows you to view user resources.
 ``users list``
 ==============
 
+.. warning::
+
+    The commands listed in this section (``users list``) are for internal use
+    by CrateDB and CrateDB Cloud staff only. They are listed here only to
+    clarify their function, since they appear in the full commands list
+    available under ``--help``.
+
 .. argparse::
    :module: croud.__main__
    :func: get_parser
--- a/croud/parser.py
+++ b/croud/parser.py
@@ -140,7 +140,7 @@ def add_default_args(parser, omit: Set[str]):
             "-o",
             required=False,
             choices=OUTPUT_FORMATS,
-            help="Change the formatting of the output",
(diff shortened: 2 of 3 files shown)


docs/commands/users.rst
```diff
@@ -2,6 +2,12 @@
 ``users``
 =========
 
+.. warning::
+
+    The commands listed in this section are for internal use by CrateDB Cloud
```
MarkusH, line 7: `$ croud users roles list` is available to any authenticated user, not just superusers.

matthijskrul: So it's not true for all subcommands there? That was not mentioned in the story card, I will change that then

mechanomi: okay, it looks as if my understanding was out-of-date when we spoke yesterday, matthijs. I'm sorry about that

looking through the documentation now, I see various places where "superuser" (or "superuser") are mentioned. I believe this term refers to Crate staff members

the term "superuser" is not defined in the documentation. see, for example, https://crate.io/docs/cloud/cli/en/latest/user-roles.html

at the very least, imo, we should add something to the docs that clearly explain that by "superuser" we mean CrateDB Cloud staff member. and uses of the word "superuser" should probably link through to this definition

for what it's worth, I don't like this use of the term superuser. I think we're stretching the meaning of that word in a way that is confusing and, more importantly, I don't think we should be exposing staff tooling to the end-user

matthijskrul: Ok, after some discussion I'll just add the warning for the relevant section for now, and document the meaning of superuser in the Croud user roles (which is otherwise anyway redundant)

PR #298, opened 2020-09-22
Remove Travis CI from project

chaudum:
Documentation is built and tested on readthedocs

## Checklist


mechanomi (changes requested):
can we please leave the docs tests in place for now

at the moment, all the RTD checks do is test that the build is working *on RTD*. as it stands, we run checks/tests that RTD does not. specifically, we run `make test` which includes things like link checking and RST linting (including natural language style guide stuff)

we should not remove the Jenkins tests without a replacement solution (such as using GitHub Actions)

as it stands: we are interested in moving all docs testing to GitHub Actions, but we haven't looked at that yet, so there hasn't been any progress

if you are able to replicate this Jenkins setup with GitHub Actions, that would be useful. because then we could use that as a template and mirror it across to the other docs projects

but one thing I don't want to do is to let the docs projects drift out of sync with each other. we're almost at a point where all of our public docs projects are in lock-step wrt to build, testing, and CI/CD. I'd like to make sure we only make forward progress on that front

what do you think?

mechanomi:
will be fixed by https://github.com/crate/croud/pull/299

PR #302, opened 2020-11-19
Support for wide tabular output format, where all fields are displayed

juanpardo:

`-o wide` can be used to print the full table as opposed to the cherry picked columns the argument `-o table` provides.

## Checklist

### DIFF
--- a/croud/organizations/auditlogs/commands.py
+++ b/croud/organizations/auditlogs/commands.py
@@ -36,7 +36,7 @@ iso8601_datetime_re = re.compile(
 )
 
 
-def actor_id_transfor(field):
+def actor_id_transform(field):
     return field["id"] or "SYSTEM"
 
 
@@ -80,5 +80,5 @@ def auditlogs_list(args: Namespace) -> None:
         errors=errors,
         keys=["action", "actor", "created"],
         output_fmt=get_output_format(args),
-        transforms=[None, actor_id_transfor, None],
+        transforms={"actor": actor_id_transform},
     )
--- a/croud/printer.py
+++ b/croud/printer.py
@@ -34,7 +34,7 @@ def print_format(
     rows: Union[List[JsonDict], JsonDict],
     format: str,
     keys: Optional[List[str]] = None,
-    transforms: List[Optional[Callable[[Any], Any]]] = None,
+    transforms: Optional[Dict[str, Callable[[Any], Any]]] = None,
 ) -> None:
     try:
         Printer = PRINTERS[format]
@@ -51,7 +51,7 @@ def print_response(
     output_fmt,
     success_message: str = None,
     keys: List[str] = None,
-    transforms: List[Optional[Callable[[Any], Any]]] = None,
+    transforms: Dict[str, Callable[[Any], Any]] = None,
 ):
     if errors:
         if "message" in errors:
@@ -94,15 +94,11 @@ class FormatPrinter(abc.ABC):
     def __init__(
         self,
         keys: Optional[List[str]] = None,
-        transforms: List[Optional[Callable[[Any], Any]]] = None,
+        transforms: Optional[Dict[str, Callable[[Any], Any]]] = None,
     ):
-        assert not transforms or transforms and keys and len(keys) == len(transforms), (
-            "When providing transforms the number of transforms must match the number "
-            "of provided keys."
-        )
 
         self.keys = keys
-        self.transforms = transforms
+        self.transforms = transforms or {}
 
     def print_rows(self, rows: Union[List[JsonDict], JsonDict]) -> None:
         print(self.format_rows(rows))
@@ -118,19 +114,8 @@ class JsonFormatPrinter(FormatPrinter):
 
 
 class TableFormatPrinter(FormatPrinter):
-    transform_funcs: Dict[str, Callable[[Any], Any]] = {}
 
-    def __init__(
-        self,
-        keys: Optional[List[str]] = None,
-        transforms: List[Optional[Callable[[Any], Any]]] = None,
-    ):
-        super().__init__(keys=keys, transforms=transforms)
-        if keys and transforms:
-            self.transform_funcs = {
-                key: transform or TableFormatPrinter._identity_transform
-                for key, transform in zip(keys, transforms)
-            }
+    display_all_columns = False
 
     def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
         if rows is None:
@@ -139,7 +124,7 @@ class TableFormatPrinter(FormatPrinter):
         if not isinstance(rows, list):
             rows = [rows]
 
-        if self.keys:
+        if self.keys and (not self.display_all_columns or len(rows) == 0):
             filter_record = functools.partial(self._filter_record, keys=self.keys)
             rows = list(map(filter_record, rows))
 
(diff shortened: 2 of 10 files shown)


MarkusH (commented):
Overall looks really good already!

croud/printer.py
```diff
@@ -94,12 +94,8 @@ class FormatPrinter(abc.ABC):
...
-        transforms: List[Optional[Callable[[Any], Any]]] = None,
+        transforms: Optional[Dict[str, Callable[[Any], Any]]] = None,
     ):
-        assert not transforms or transforms and keys and len(keys) == len(transforms), (
-            "When providing transforms the number of transforms must match the number "
-            "of provided keys."
-        )
 
         self.keys = keys
         self.transforms = transforms
```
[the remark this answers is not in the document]
juanpardo: Thanks, I didn't know that was possible, it's great syntactic sugar!

croud/printer.py
```diff
@@ -118,19 +114,17 @@ def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
 
 
 class TableFormatPrinter(FormatPrinter):
-    transform_funcs: Dict[str, Callable[[Any], Any]] = {}
+
+    display_all_columns: Optional[bool] = False
```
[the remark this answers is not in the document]
juanpardo: It worked well for normal cases, but then the tests failed at the time of displaying zero result data, where it was printing nothing. After some thought with @christianbader we opted for `wide` to provide the same result as `table` when there are no results to display. That means wiping the keys is not a valid solution anymore so I moved the complexity into the `TableFormatPrinter` and left the `WideTableFormatPrinter` as a configuration class that fully delegates this logic into its parent class.

Given this scenario I think Option 1 suits it better but very curious to read further thoughts on this :)

croud/printer.py
```diff
@@ -118,19 +114,15 @@ def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
...
+
+    display_all_columns: bool = False
 
     def __init__(
         self,
         keys: Optional[List[str]] = None,
-        transforms: List[Optional[Callable[[Any], Any]]] = None,
+        transforms: Optional[Dict[str, Callable[[Any], Any]]] = None,
     ):
         super().__init__(keys=keys, transforms=transforms)
```
MarkusH, line 125: Since you don't do anything but call `super().__init__(...)` in this method, you don't need the method at all :)

croud/printer.py
```diff
@@ -185,6 +178,22 @@ def _identity_transform(field):
...
 
+class WideTableFormatPrinter(TableFormatPrinter):
+    """
+    Prints data in tabular format without any field restrictions, displaying all the
+    properties of the payload.
+    """
+
+    display_all_columns: bool = True
+
+    def __init__(
```
MarkusH, line 189: As for the `TableFormatPrinter`, you don't need the `__init__()` method at all, since you're only calling the super class' method.

MarkusH (commented):
Two small cleanups. Looks great!

croud/printer.py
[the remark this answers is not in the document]
MarkusH: Good point!

mechanomi (changes requested):
please can you update the corresponding parts of the documentation?

e.g., https://github.com/crate/croud/blob/master/docs/configuration.rst#configuration-file-keys

croud/printer.py
```diff
@@ -118,19 +114,8 @@ def format_rows(self, rows: Union[List[JsonDict], JsonDict]) -> str:
...
-        keys: Optional[List[str]] = None,
-        transforms: List[Optional[Callable[[Any], Any]]] = None,
-    ):
-        super().__init__(keys=keys, transforms=transforms)
-        if keys and transforms:
-            self.transform_funcs = {
-                key: transform or TableFormatPrinter._identity_transform
-                for key, transform in zip(keys, transforms)
-            }
+    display_all_columns: bool = False
```
MarkusH, line 118: Sorry, that I missed this during my last review. You don't need a type annotation here, since mypy can derive it from the variable value.

croud/printer.py
```diff
@@ -185,6 +171,15 @@ def _identity_transform(field):
             return field
 
 
+class WideTableFormatPrinter(TableFormatPrinter):
+    """
+    Prints data in tabular format without any field restrictions, displaying all the
+    properties of the payload.
+    """
+
+    display_all_columns: bool = True
```
MarkusH, line 180: As above.

MarkusH (approved):

juanpardo:
> please can you update the corresponding parts of the documentation?
> 
> e.g., https://github.com/crate/croud/blob/master/docs/configuration.rst#configuration-file-keys

@mechanomi thanks for your feedback.

I got a question on how to introduce this in the documentation, at the moment there is no mention on the argument `-o` at all anywhere in the documentation, but it already existed before this pull request (with options `table`, `json` and `yaml`).
There is a mention of the options on the default-format part of the documentation in `configuration.rst`. I was wondering if in addition to add it to the supported list of default formats I should also add an entry somewhere to describe what the `-o`  argument does. If so, could you give me an example, even from another project? And where should I put this :P

mechanomi:
yep, good catch, @juanpardo 

I think the following section needs to be updated: https://github.com/crate/croud/blob/master/docs/configuration.rst#configuration-file-keys

the `-o` option appears to be documented in multiple places, all of them in the commands section of the docs. see, for example: https://crate.io/docs/cloud/cli/en/latest/commands/config.html

search "Optional Arguments"

so, all of the "Possible choices" lists that appears every time `-o` is mentioned need updating too

juanpardo:
> please can you update the corresponding parts of the documentation?
> 
> e.g., https://github.com/crate/croud/blob/master/docs/configuration.rst#configuration-file-keys

Added the documentation bits. Asked Daisy for help on this area. Turns out the other parts of the docs are generated dynamically so it should be sufficient.

croud/organizations/auditlogs/commands.py
```diff
@@ -80,5 +80,5 @@ def auditlogs_list(args: Namespace) -> None:
         errors=errors,
         keys=["action", "actor", "created"],
         output_fmt=get_output_format(args),
-        transforms=[None, actor_id_transfor, None],
+        transforms={"actor": actor_id_transfor},
```
chaudum, line 83: Would you mind fixing the name of the transform method?

```suggestion
        transforms={"actor": actor_id_transform},
```

chaudum (approved):
Otherwise looks great!

mechanomi:
@juanpardo ah. interesting. I did not know that. lgtm then

mechanomi (approved):

croud/organizations/auditlogs/commands.py
juanpardo: Nice catch!

docs/configuration.rst
```diff
@@ -51,7 +51,7 @@ The keys have the following meaning:
 :``default-format``:
 
     The default output format for API requests. This can be either ``table``
-    (default), ``json``, or ``yaml``. This value is only used if the current
+    (default, displays the most important fields), ``wide`` (a table format displaying all fields), ``json``, or ``yaml``. This value is only used if the current
```
MarkusH, line 54: I'd add line breaks to follow the style of the surrounding docs.

MarkusH (approved):

chaudum:
:tada:

PR #307, opened 2020-12-04
Added an (optional) region param

SStorm:

This allows us to set a default region (which can also be __any__) and not have to ever specify one when running any commands.

## Checklist

### DIFF
--- a/croud/api.py
+++ b/croud/api.py
@@ -93,7 +93,7 @@ class Client:
             CONFIG.endpoint,
             token=CONFIG.token,
             on_token=CONFIG.set_current_auth_token,
-            region=args.region if args.region is not None else CONFIG.region,
+            region=args.region or CONFIG.region,
             sudo=args.sudo,
         )
 
--- a/croud/api.py
+++ b/croud/api.py
@@ -93,7 +93,7 @@ class Client:
             CONFIG.endpoint,
             token=CONFIG.token,
             on_token=CONFIG.set_current_auth_token,
-            region=args.region,
+            region=args.region if args.region is not None else CONFIG.region,
             sudo=args.sudo,
         )
 
--- a/docs/commands/regions.rst
+++ b/docs/commands/regions.rst
@@ -15,7 +15,9 @@ Print the available regions to the user:
 
 .. note::
 
-   Listed region names can used in the ``--region`` argument to list/deploy resources in that region.
+   The region is specified for each profile in the :doc:`../configuration` file.
+   The region for specific actions can be overridden using the ``--region`` argument to list/deploy resources in that region.
+
 
 Example
 =======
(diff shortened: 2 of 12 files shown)


chaudum:
I like the change a lot! It will make the distinction between endpoint and region clearer.

However, I would only put not deprecated regions into the default profile: 

```yaml
default-format: table
current-profile: aks1.eastus2.azure
profiles:
  aks1.eastus2.azure:
    auth-token: NULL
    endpoint: https://console.cratedb.cloud
    region: aks1.eastus2.azure
  aks1.westeurope.azure:
    auth-token: NULL
    endpoint: https://console.cratedb.cloud
    region: aks1.westeurope.azure
  eks1.eu-west-1.aws:
    auth-token: NULL
    endpoint: https://console.cratedb.cloud
    region: eks1.eu-west-1.aws
```

Additionally, we need to update the documentation a bit, specifically [docs/configuration.rst](https://github.com/crate/croud/blob/master/docs/configuration.rst) and [docs/commands/regions.rst](https://github.com/crate/croud/blob/master/docs/commands/regions.rst). Maybe you want to check with @matthijskrul how we could pimp the information on regions.

SStorm:
Have updated the docs and the profiles list to only show the non-deprecated ones. @matthijskrul docs feedback will be highly appreciated!

croud/api.py
```diff
@@ -93,7 +93,7 @@ def from_args(args: Namespace) -> "Client":
             CONFIG.endpoint,
             token=CONFIG.token,
             on_token=CONFIG.set_current_auth_token,
-            region=args.region,
+            region=args.region if args.region is not None else CONFIG.region,
```
chaudum, line 96: If you don't want to explicitly check for `None`, you could simpler write:

```suggestion
            region=args.region or CONFIG.region,
```

chaudum (approved):
Great stuff!

docs/configuration.rst
```diff
@@ -59,12 +59,14 @@ The keys have the following meaning:
...
     (see :ref:`available-profiles`).
-    Each profile consists of ``auth-token``, ``endpoint``, and ``format``.
+    Each profile consists of ``auth-token``, ``endpoint``, ``region``, and ``format``.
 
     ``auth-token`` is populated with the API token upon login.
 
     ``endpoint`` is the full URL of the API endpoint that is used for requests.
 
+    ``region`` is the specific region where CrateDB Cloud resources are accessed in.
+
```
matthijskrul, line 69: I'd rephrase this: something like "``region`` is the specific region within which CrateDB Cloud resources are accessed".

docs/configuration.rst
```diff
@@ -84,13 +86,15 @@ details.
...
 ==================
 
-==================== ====================================== ===========
-Profile              Endpoint                               Format
-==================== ====================================== ===========
-bregenz.a1           https://bregenz.a1.cratedb.cloud       table
-eastus2.azure        https://eastus2.azure.cratedb.cloud    table
-westeurope.azure     https://westeurope.azure.cratedb.cloud table
-==================== ====================================== ===========
+The API Endpoint for all profiles is ``https://console.cratedb.cloud``
```
matthijskrul, line 89: Should have a period at the end. Also I'd do lowercase "endpoint"

docs/commands/regions.rst
```diff
@@ -15,19 +15,21 @@ Print the available regions to the user:
 
 .. note::
 
-   Listed region names can used in the ``--region`` argument to list/deploy resources in that region.
+   The region is specified for each profile in the :doc:`../configuration` file.
+   The region for specific actions can be overridden using the ``--region`` argument to list/deploy resources in that region.
```
matthijskrul, line 19: Preferably "list or deploy"

PR #309, opened 2020-12-10
Specify a custom User-Agent

SStorm:

Will make it more distinguishable in the logs.

## Checklist

### DIFF
--- a/croud/api.py
+++ b/croud/api.py
@@ -19,11 +19,13 @@
 
 import enum
 from argparse import Namespace
+from platform import python_version
 from typing import Any, Callable, Dict, Optional, Tuple
 
 import requests
 from yarl import URL
 
+import croud
 from croud.config import CONFIG
 from croud.printer import print_error, print_info
 
@@ -86,6 +88,9 @@ class Client:
             self.session.headers["X-Region"] = region
         if sudo:
             self.session.headers["X-Auth-Sudo"] = "1"
+
+        ua = f"Croud/{croud.__version__} Python/{python_version()}"
+        self.session.headers["User-Agent"] = ua
 
     @staticmethod
     def from_args(args: Namespace) -> "Client":
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,9 @@ Unreleased
   This makes it explicit which one is being used as the recommended API endpoint
   is always ``https://console.cratedb.cloud``.
 
+- Croud will now send a custom User-Agent including the version of ``croud`` itself
+  and the version of the Python interpreter that is running it.
+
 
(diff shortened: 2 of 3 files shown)


croud/api.py
```diff
@@ -77,6 +78,7 @@ def __init__(
         self.base_url = URL(endpoint)
         self._token = token
         self._on_token = on_token or noop
+        self._croud_version = get_distribution("croud").version
```
chaudum, line 81: You could use the version from `croud.__version__`.

croud/api.py
```diff
@@ -116,9 +118,14 @@ def request(
...
 
+        headers = {"User-Agent": f"Croud {self._croud_version}"}
+
         try:
             response = self.session.request(
-                method.value, str(self.base_url.with_path(endpoint)), **kwargs
+                method.value,
+                str(self.base_url.with_path(endpoint)),
+                headers=headers,
+                **kwargs,
```
chaudum, line 128: I think it is more consistent with other headers if the `User-Agent` header is set on the session:

https://github.com/crate/croud/blob/master/croud/api.py#L81

croud/api.py
```diff
@@ -116,9 +118,14 @@ def request(
         if body is not None:
             kwargs["json"] = body
 
+        headers = {"User-Agent": f"Croud {self._croud_version}"}
```
chaudum, line 121: I would recommend to follow the proposed anatomy of a User-Agent string as described in [RFC7231](https://tools.ietf.org/html/rfc7231#section-5.5.3).

Either simple: `Croud/{version}` or even `Croud/{croud_version} Python/{py_version}` or with additional "comments" about the operating system `Croud/{version} (Linux; ...)`

croud/api.py
```diff
@@ -87,6 +89,10 @@ def __init__(
         if sudo:
             self.session.headers["X-Auth-Sudo"] = "1"
 
+        self.session.headers[
+            "User-Agent"
+        ] = f"Croud/{croud.__version__} Python/{python_version()}"
```
chaudum, line 94: I am aware that `black` does this sort of ugly formatting automatically. I guess if you'd split up into two lines, it would be more readable.

chaudum (commented):
Other than that, lgtm! Please also add a changelog entry.

croud/api.py
SStorm: Yeah I tried it several different ways and it kept insisting on doing it this way :) (n.b. I like `black` - it eliminates useless discussions about trivial code style issues that would be happening if we weren't enforcing it this way). Have split it into two lines now and it's better.

chaudum (approved):

PR #310, opened 2020-12-11
Get resources by their ID

chaudum:

This commit adds a new subcommand `get` for the following resources:

* `clusters`
* `organizations`
* `projects`
* `subscriptions`

The command requires a single positional argument `ID` and it returns a
single item of the requested resource by fetching the item by its ID
using the `GET /api/v2/{resource}/{id}/` API endpoints.

Example:

```console
$ croud clusters get 205815c3-230f-442d-99e8-1c34bd717845
...
```

### DIFF
--- a/croud/clusters/commands.py
+++ b/croud/clusters/commands.py
@@ -25,6 +25,18 @@ from croud.printer import print_response
 from croud.util import require_confirmation
 
 
+def clusters_get(args: Namespace) -> None:
+    """
+    Get cluster by ID
+    """
+
+    client = Client.from_args(args)
+    data, errors = client.get(f"/api/v2/clusters/{args.id}/")
+    print_response(
+        data=data, errors=errors, output_fmt=get_output_format(args),
+    )
+
+
 def clusters_list(args: Namespace) -> None:
     """
     Lists all projects for the current user in the specified region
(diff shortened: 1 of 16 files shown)


croud/clusters/commands.py
```diff
@@ -25,6 +25,18 @@
 from croud.util import require_confirmation
 
 
+def clusters_get(args: Namespace) -> None:
+    """
+    Get cluster by ID
```
turbo-ele, line 30:
```suggestion
    Get cluster by ID.
```
as far as i've seen, isn't that the convention ?

turbo-ele (commented):
Changes entry missing?

SStorm (approved):
Works nicely!

croud/clusters/commands.py
chaudum: Makes sense (even we don't follow the convention everywhere).

chaudum: Actually, I don't think the doc strings add much to the commands (anywhere).

chaudum: @turbo-ele Removed them in a new commit.

turbo-ele (approved):
lgtm

PR #313, opened 2021-02-11
Add subcommand to fetch a deployment manifest for an edge region

lukasbals:

Add subcommand to fetch a deployment manifest for an edge region.

On hold since the backend endpoint is not yet released.

## Checklist

### DIFF
--- a/croud/regions/commands.py
+++ b/croud/regions/commands.py
@@ -18,10 +18,11 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 from argparse import Namespace
+from base64 import b64decode
 
 from croud.api import Client
 from croud.config import get_output_format
-from croud.printer import print_response
+from croud.printer import print_error, print_response, print_success
 
 
 def regions_list(args: Namespace) -> None:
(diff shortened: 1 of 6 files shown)


croud/regions/commands.py
```diff
@@ -37,3 +38,28 @@ def regions_list(args: Namespace) -> None:
...
+    """
+
+    client = Client.from_args(args)
+    print_info("Generating deployment manifest ...")
+    data, errors = client.get(
+        f"/api/v2/regions/{args.region_name}/deployment-manifest/"
+    )
+    if data:
+        try:
+            f = open(data["name"], "x")
```
SStorm, line 56: Whoa, totally not obvious that it's going to write to a file here... Would it be better to print this to `stdout` and let the user of `croud` redirect it where he wants to?

SStorm: I mean, what if the server returns '/etc/passwd' and I'm running croud as root? :D

lukasbals: Good point. Working on a solution.

croud/regions/commands.py
```diff
@@ -37,3 +38,32 @@ def regions_list(args: Namespace) -> None:
...
+
+    client = Client.from_args(args)
+    data, errors = client.get(
+        f"/api/v2/regions/{args.region_name}/deployment-manifest/"
+    )
+    if data:
+        content = b64decode(data["content"]).decode()
+        if args.file_name:
+            try:
+                f = open(args.file_name, "x")
```
SStorm, line 57: You could use `with open(...) as file:` I suppose :shrug:

SStorm (approved):
LGTM!

croud/regions/commands.py
```diff
@@ -37,3 +38,32 @@ def regions_list(args: Namespace) -> None:
...
+    data, errors = client.get(
+        f"/api/v2/regions/{args.region_name}/deployment-manifest/"
+    )
+    if data:
+        content = b64decode(data["content"]).decode()
+        if args.file_name:
+            try:
+                with open(args.file_name, "x") as file:
+                    file.write(content)
+                    file.close()
```
SStorm, line 59: You don't need `file.close()` if you use `with ...` :stuck_out_tongue:

matthijskrul (approved):

lukasbals:
I will close this PR since there has a feature branch been created: https://github.com/crate/croud/pull/316

I will create a new PR against the feature branch: `feature/bootstrap-edge`

PR #315, opened 2021-02-15
Fix loading indicator to also disappear if just a message is printed

lukasbals:

Fix loading indicator also to disappear if just a message is printed.

To avoid something like that, where the loading indicator doesn't disappear after the command has finished:

Screenshot 2021-02-15 at 11 29 22

## Checklist


SStorm (approved):
Good find!

PR #326, opened 2021-04-13
Corrected install edge-region script command when creating a new region

turbo-ele:

- Corrected install edge-region script command when creating a new region
- Added a new raw format printer
- Added a test for the returned install command
- Changed the command to use `wget` instead of curl

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Added a new type of printer ``raw`` that simply outputs the rows as given.
+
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Added new type of printer `raw` that simply outputs the rows as given.
```
lukasbals, line 8:
```suggestion
- Added a new type of printer ``raw`` that simply outputs the rows as given.
```

lukasbals (approved):

PR #327, opened 2021-04-14
Fixed linter errors. Removed RawFormatPrinter

turbo-ele:

.. as formatprinters only support JsonDict flavors and not simple strings.


lukasbals (approved):
LGTM 🙂

PR #332, opened 2021-06-01
Allow specifying custom hardware specs for edge products

SStorm:

Note that this requires the corresponding API changes to be merged.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,9 +61,6 @@ coverage.xml
 # PyBuilder
 target/
 
-# buildout
-env/
-
 # vscode
 .vscode/
 .vscode/*
--- a/tests/commands/test_clusters.py
+++ b/tests/commands/test_clusters.py
@@ -96,6 +96,65 @@ def test_clusers_deploy(mock_request):
 
 
 @mock.patch.object(Client, "request", return_value=({}, None))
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -61,9 +61,6 @@ coverage.xml
 # PyBuilder
 target/
 
-# buildout
```
SStorm, line 64: This was a duplicate.

tests/commands/test_clusters.py
```diff
@@ -95,6 +95,65 @@ def test_clusers_deploy(mock_request):
...
+        "premium",
+        "--memory-size-mb",
+        "2048",
+    )
+    assert_rest(
+        mock_request,
+        RequestMethod.POST,
+        "/api/v2/clusters/",
+        body={
+            "crate_version": "3.2.5",
```
juanpardo, line 137: Shouldn't this be 4.5.1 too?

juanpardo (approved):

tests/commands/test_clusters.py
SStorm: These are just in the tests, doesn't really matter which version it is I suppose?

I've updated the other ones as they are in the public docs and we probably want to advertise a more recent version of Crate!

PR #337, opened 2021-07-13
check region status before deleting it

plaharanne:
## Checklist

### DIFF
--- a/croud/regions/commands.py
+++ b/croud/regions/commands.py
@@ -21,7 +21,7 @@ from argparse import Namespace
 
 from croud.api import Client
 from croud.config import get_output_format
-from croud.printer import print_raw, print_response, print_success
+from croud.printer import print_error, print_raw, print_response, print_success
 from croud.util import require_confirmation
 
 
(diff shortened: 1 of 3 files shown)


croud/regions/commands.py
```diff
@@ -110,6 +110,25 @@ def regions_delete(args: Namespace) -> None:
     client = Client.from_args(args)
     region_name = args.name
 
+    # Check edge region status
+    regions, errors = client.get("/api/v2/regions/")
```
SStorm, line 114: Just realised we don't actually have a `GET /api/v2/regions/<id>` call :facepalm:

croud/regions/commands.py
```diff
@@ -110,6 +110,25 @@ def regions_delete(args: Namespace) -> None:
...
+    region = [r for r in (regions or []) if r["name"] in region_name]
```
SStorm, line 115: Could maybe do something like this:
```suggestion
    region = next((r for r in (regions or []) if r["name"] == region_name), None)
```
? 
Would immediately get the region or None, and could in fact immediately short-circuit and do a
```
if not region:
    raise ...
```

SStorm (approved):
Nice work! Added a couple of suggestions if you feel like it.

SStorm (approved):
Well done!

PR #341, opened 2021-08-10
Allow specifying an S3 backup-location-endpoint-url for projects 

WalBeh:
Adding `--backup-location-endpoint-url` allows specifying a custom S3 endpoint (eg. http://minio.svc.local:9000) for cluster backup. This setting is on `project` level.

## Checklist

### DIFF
--- a/croud/projects/commands.py
+++ b/croud/projects/commands.py
@@ -109,6 +109,10 @@ def _handle_custom_backups(body, args: Namespace) -> None:
             "access_key_id": args.backup_location_access_key_id,
             "secret_access_key": args.backup_location_secret_access_key,
         }
+    if args.backup_location_endpoint_url:
+        body.setdefault("backup_location", {})["additional_config"] = {
+            "endpoint_url": args.backup_location_endpoint_url,
+        }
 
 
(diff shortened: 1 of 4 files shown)


croud/projects/commands.py
```diff
@@ -109,6 +109,10 @@ def _handle_custom_backups(body, args: Namespace) -> None:
             "access_key_id": args.backup_location_access_key_id,
             "secret_access_key": args.backup_location_secret_access_key,
         }
+    if args.backup_location_endpoint_url:
+        body.setdefault("backup_location", {})[
```
juanpardo, line 113: I am afraid this will not work since in the API I implemented ``endpoint_url`` inside another category, at the same level than ``credentials``, called ``additional_config``.
The reasoning behind this was to have all non-generic parameters (that is, the ones that depend on the backup_location_type) either under ``credentials`` (if they must not be persisted but still passed on) or ``additional_config`` (if they should be persisted in Brain).

juanpardo (changes requested):

juanpardo (approved):
Looks great now!

PR #342, opened 2021-08-12
Add new command to set the deletion protection status of a cluster

tomach:

Added new subcommand ``clusters set-deletion-protection`` that allows superusers and organization admins to set the deletion protection status of a cluster.

## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -20,6 +20,7 @@
 # software solely pursuant to the terms of the relevant commercial agreement.
 
 import sys
+from distutils.util import strtobool
 
 import colorama
 
@@ -30,6 +31,7 @@ from croud.clusters.commands import (
     clusters_list,
     clusters_restart_node,
     clusters_scale,
+    clusters_set_deletion_protection,
     clusters_upgrade,
 )
 from croud.config import CONFIG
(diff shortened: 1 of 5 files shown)


croud/__main__.py
```diff
@@ -457,6 +459,20 @@
...
             },
+            "set-deletion-protection": {
+                "help": "Set the deletion protection status of a CrateDB cluster.",
+                "extra_args": [
+                    Argument(
+                        "--cluster-id", type=str, required=True,
+                        help="The CrateDB cluster ID to use.",
+                    ),
+                    Argument(
+                        "--value", type=lambda x: bool(strtobool(str(x))),
```
SStorm, line 470: Interesting lambda for the type here, do we do this somewhere else yet? Could we use (is it supported) just `type=bool`?..

SStorm (approved):
Nice work!

croud/__main__.py
tomach: I don't think we use it somewhere else, at least I didn't find it.. `type=bool` would be allowed but it does not behave as expected, the `type` is a callable which is used as `argument = type(argument)` but  `bool('False')` evaluates to True unfortunately :)

PR #354, opened 2021-12-06
Some tweaks to the status reporting of upgrades

SStorm:

- Will not print the whole response table but individual statuses
- Will not print statuses if they have not changed

Screenshot from 2021-12-06 13-30-15

## Checklist


tomach (approved):
looks great :+1:

PR #358, opened 2022-01-21
Add status checks to cluster creation and ip-whitelist update

tomach:
Use the same status reporting for cluster creation and ip-whitelist updates as it is already used for cluster scaling and upgrades.

https://github.com/crate/cloud/issues/439

image

## Checklist

 - [X] [CLA](https://crate.io/community/contribute/cla/) is signed

### DIFF
--- a/croud/clusters/commands.py
+++ b/croud/clusters/commands.py
@@ -80,9 +80,26 @@ def clusters_deploy(args: Namespace) -> None:
         data=data,
         errors=errors,
         keys=["id", "name", "fqdn", "url"],
-        success_message=(
-            "Cluster deployed. It may take a few minutes to complete the changes."
-        ),
+        output_fmt=get_output_format(args),
+    )
+
+    if errors or not data:
+        return
+
+    print_info("Cluster creation initiated. It may take a few minutes to complete.")
+
+    _wait_for_completed_operation(
+        client=client,
+        cluster_id=data["id"],
+        request_params={"type": "CREATE", "limit": 1},
+    )
+
+    # Re-fetch the cluster's info
+    data, errors = client.get(f"/api/v2/clusters/{data['id']}/")
+    print_response(
+        data=data,
+        errors=errors,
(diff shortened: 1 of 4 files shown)


croud/clusters/commands.py
```diff
@@ -80,9 +80,53 @@ def clusters_deploy(args: Namespace) -> None:
...
-        ),
+        output_fmt=get_output_format(args),
+    )
+
+    if errors or not data:
+        return
+
+    print_info(
+        "Cluster creation initiated. "
+        "It may take a few minutes to complete the changes."
```
SStorm, line 91:
```suggestion
        "It may take a few minutes to complete."
```

croud/clusters/commands.py
```diff
@@ -80,9 +80,53 @@ def clusters_deploy(args: Namespace) -> None:
...
+    )
+    params = {"type": "CREATE", "limit": 1}
+    last_status = None
+    last_msg = None
+    while True:
```
SStorm, line 96: Now that this is happening in at least 4 places, I think we definitely want a helper method for it! :) Some sort of

```python
wait_for_completed_operation("cluster", data['id'], params={"type": "CREATE", "limit": 1})
```

croud/clusters/commands.py
```diff
@@ -80,9 +80,53 @@ def clusters_deploy(args: Namespace) -> None:
...
+        try:
+            status, msg = _get_operation_status(
+                client=client, cluster_id=data["id"], request_params=params
+            )
+        except AsyncOperationNotFound as e:
+            print_error(str(e))
+            break
+
+        if status == "SUCCEEDED":
+            print_success("Cluster successfully created.")
```
SStorm, line 106: If we printed "Operation completed" this could be a generic piece of code I think?

SStorm (commented):
What do you say about generifying the waiting of the results?

SStorm (approved):
:partying_face:

PR #368, opened 2022-04-28
Use the new cluster provisioning API

SStorm:

Depends on https://github.com/crate/cloud-api/pull/2117

## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -419,7 +419,13 @@ command_tree = {
                         "--memory-size-mb", type=int, required=False,
                         help="Amount of memory to allocate (in MiB). "
                              "CrateDB Edge regions only.",
-                    )
+                    ),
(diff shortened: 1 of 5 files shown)


croud/__main__.py
```diff
@@ -420,7 +420,13 @@
                         "--memory-size-mb", type=int, required=False,
                         help="Amount of memory to allocate (in MiB). "
                              "CrateDB Edge regions only.",
-                    )
+                    ),
+                    Argument(
+                        "--subscription-id", type=str,
```
tomach, line 425: I think we should document this argument also in `docs/commands/clusters.rst`

tomach: And remove the sentence that `clusters deploy` is for superusers only there.

tomach (approved):
:+1:

PR #371, opened 2022-06-03
Add new command to suspend or resume a cluster

tomach:

Added new subcommand `clusters set-suspended-state` that allows superusers, organization admins and project admins to suspend or resume a cluster.

## Checklist

### DIFF
--- a/croud/clusters/commands.py
+++ b/croud/clusters/commands.py
@@ -239,6 +239,40 @@ def clusters_set_deletion_protection(args: Namespace) -> None:
     )
 
 
+def clusters_set_suspended(args: Namespace) -> None:
+    body = {"suspended": args.value}
+    client = Client.from_args(args)
+    data, errors = client.put(f"/api/v2/clusters/{args.cluster_id}/suspend/", body=body)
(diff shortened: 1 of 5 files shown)


croud/clusters/commands.py
```diff
@@ -239,6 +239,40 @@ def clusters_set_deletion_protection(args: Namespace) -> None:
...
+    data, errors = client.put(f"/api/v2/clusters/{args.cluster_id}/suspend/", body=body)
+    print_response(
+        data=data,
+        errors=errors,
+        keys=["id", "name", "suspended"],
+        output_fmt=get_output_format(args),
+    )
+
+    if errors:
+        return
```
plaharanne, line 254: you don't print any error message if it fails?

plaharanne (approved):

croud/clusters/commands.py
tomach: I think the errors are already printed in `print_response` [here](https://github.com/crate/croud/blob/master/croud/printer.py#L57) if it fails

PR #376, opened 2022-07-21
Update regions create command

Taliik:
## Checklist


tomach (approved):
:100: Is there anything in the docs that needs to be updated here https://github.com/crate/croud/blob/f/update-regions-commands/docs/commands/regions.rst? I think only the `--provider` argument should be removed from the example, right?

PR #387, opened 2022-11-14
Add support to changes the backup schedule of clusters

juanpardo:
## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -571,7 +571,7 @@ command_tree = {
                         "--backup-hours", type=str, required=True,
                         help="A list of hours in UTC time that represent when backups "
                              "will take place along the day. The list must contain at "
-                             "least one value and can have up 23, comma-separated."
+                             "least one value and can have up to 24, comma-separated."
                     )
                 ],
                 "resolver": clusters_change_backup_schedule,
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -26,6 +26,7 @@ import colorama
 import shtab
 
 from croud.clusters.commands import (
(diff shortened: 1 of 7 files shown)


croud/clusters/commands.py
```diff
@@ -403,6 +403,23 @@ def clusters_change_product(args: Namespace) -> None:
     )
 
 
+def clusters_change_backup_schedule(args: Namespace) -> None:
+    body = {
+        "backup_hours": args.backup_hours,
+    }
+
+    client = Client.from_args(args)
```
SStorm, line 411: As this starts an async operation, would it make sense to wait for it to complete, like the other async methods do? i.e. see above in the `clusters_change_product` method.

croud/__main__.py
```diff
@@ -558,6 +559,22 @@
...
+                "extra_args": [
+                    Argument(
+                        "--cluster-id", type=str, required=True,
+                        help="The CrateDB cluster ID to use.",
+                    ),
+                    Argument(
+                        "--backup-hours", type=str, required=True,
+                        help="A list of hours in UTC time that represent when backups "
+                             "will take place along the day. The list must contain at "
+                             "least one value and can have up 23, comma-separated."
```
SStorm, line 573: Isn't it 24? Also just thinking, do we allow resetting the schedule back to every hour, meaning a param of `*`?

croud/clusters/commands.py
juanpardo: Added code to handle this, thank you for the feedback :)

SStorm (approved):

PR #391, opened 2022-11-18
Added new command to cancel Stripe subscriptions

SStorm:

Also updated subscriptions list to show the org_id, as it's useful if the user has multiple orgs.

And fixed the LOG_API param to correctly use Python booleans.

## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -80,6 +80,7 @@ from croud.projects.users.commands import (
 )
 from croud.regions.commands import regions_create, regions_delete, regions_list
 from croud.subscriptions.commands import (
+    subscription_delete,
     subscriptions_create,
     subscriptions_get,
     subscriptions_list,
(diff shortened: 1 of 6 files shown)


croud/__main__.py
```diff
@@ -828,6 +829,19 @@
                 ],
                 "resolver": subscriptions_get,
             },
+            "delete": {
+                "help": "For Stripe only. Cancels the specified subscription. "
+                        "CAVEAT EMPTOR! "
+                        "This will delete any clusters running in this subscription.",
+                "extra_args": [
+                    Argument(
+                        "id", type=str,
```
tomach, line 838: Should we call the arg `--subscription-id` to be in line with the other delete commands for projects, clusters, orgs etc? And maybe also update the docs for `subscriptions list` and add this new `delete`?

tomach (approved):
:+1:

PR #394, opened 2022-11-18
Added support for Dependabot

juanpardo:
## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,22 @@
+
+version: 2
+updates:
+- package-ecosystem: pip # See documentation for possible values
+  directory: "/" # Location of package manifests
+  schedule:
+     interval: monthly
+  open-pull-requests-limit: 10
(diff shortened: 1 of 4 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,22 @@
+
```
plaharanne, line 1: There is also an other commit for that, which one is the good one?
https://github.com/crate/croud/pull/393/files

juanpardo: Since the other PR required a rebase, I decided to pick this one and introduce @SStorm 's additions and added him as a co-author. If everyone is happy with this we can go ahead with this PR and drop the other one.
Thanks you for noticing and reviewing, @plaharanne !

SStorm (approved):

.github/dependabot.yml
SStorm: Go ahead!

PR #408, opened 2022-11-29
Added support for API keys.

SStorm:

Keys / secret can now be specified in the profile instead of the auth-token. If both are present, the token takes precedence.

## Checklist


docs/configuration.rst
```diff
@@ -61,19 +55,30 @@ The keys have the following meaning:
 
 :``profiles``:
 
-    A dictionary of available profiles. There are three default profiles available
-    (see :ref:`available-profiles`).
-    Each profile consists of ``auth-token``, ``endpoint``, ``region``, and ``format``.
+    A dictionary of available profiles. There is just one profile configured in the
+    default configuration file. You would only need different profiles if you want to
+    authenticated as different users or use different default organisations.
```
tomach, line 60:
```suggestion
    authenticate as different users or use different default organizations.
```

tomach (approved):
LGTM! :100:

PR #412, opened 2022-12-09
Added support to list and restore snapshots

juanpardo:
Added support to list and restore snapshots.
This include support for restoring snapshots from a different cluster from within the organization.

## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -618,7 +618,7 @@ command_tree = {
                                 "--source-cluster-id", type=str, required=False,
                                 help="The CrateDB cluster ID of the snapshot to be "
                                      "used belongs to. Must belong to the same "
-                                     "organization than the target cluster."
+                                     "organization as the target cluster."
                                      "If not specified the ``--cluster-id`` CrateDB"
                                      " cluster will be used as the source.",
                             ),
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -602,7 +602,8 @@ command_tree = {
                         "extra_args": [
                             Argument(
                                 "--cluster-id", type=str, required=True,
-                                help="The CrateDB cluster ID to use.",
+                                help="The CrateDB cluster to restore "
+                                     "the snapshot into.",
                             ),
                             Argument(
                                 "--snapshot", type=str, required=True,
@@ -619,7 +620,7 @@ command_tree = {
                                      "used belongs to. Must belong to the same "
                                      "organization than the target cluster."
                                      "If not specified the ``--cluster-id`` CrateDB"
-                                     " cluster will be used as the target.",
+                                     " cluster will be used as the source.",
                             ),
                             Argument(
                                 "--tables", type=str, required=False,
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -44,6 +44,8 @@ from croud.clusters.commands import (
     clusters_set_ip_whitelist,
     clusters_set_product,
     clusters_set_suspended,
+    clusters_snapshots_list,
+    clusters_snapshots_restore,
     clusters_upgrade,
 )
 from croud.config import CONFIG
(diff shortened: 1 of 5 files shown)


croud/__main__.py
```diff
@@ -582,6 +584,53 @@
...
+                            ),
+                        ],
+                        "resolver": clusters_snapshots_list,
+                    },
+                    "restore": {
+                        "help": "Restore the specified snapshot.",
+                        "extra_args": [
+                            Argument(
+                                "--cluster-id", type=str, required=True,
+                                help="The CrateDB cluster ID to use.",
```
SStorm, line 605:
```suggestion
                                help="The CrateDB cluster to restore the snapshot into.",
```

croud/clusters/commands.py
```diff
@@ -442,6 +442,54 @@ def clusters_set_backup_schedule(args: Namespace) -> None:
...
+            "repository",
+            "snapshot",
+            "tables",
+            "concrete_indices",
+        ],
+        output_fmt=get_output_format(args),
+    )
+
+
+def clusters_snapshots_restore(args: Namespace) -> None:
```
SStorm, line 464: As this is an async operation, it would be good to also show progress, like with the others :)

croud/__main__.py
```diff
@@ -582,6 +584,53 @@
...
+                                help="The repository that contains the snapshot to be "
+                                     "restored.",
+                            ),
+                            Argument(
+                                "--source-cluster-id", type=str, required=False,
+                                help="The CrateDB cluster ID of the snapshot to be "
+                                     "used belongs to. Must belong to the same "
+                                     "organization than the target cluster."
+                                     "If not specified the ``--cluster-id`` CrateDB"
+                                     " cluster will be used as the target.",
```
plaharanne, line 622:
```suggestion
                                     " cluster will be used as the source.",
```

plaharanne: Unless I misunderstood

croud/clusters/commands.py
```diff
@@ -476,6 +476,19 @@ def clusters_snapshots_restore(args: Namespace) -> None:
 
     client = Client.from_args(args)
     data, errors = client.post(url, body=body)
+
+    if errors:
+        return
+    print_info(
+        "Restoring the snapshot. It may take a few minutes to complete the changes."
```
SStorm, line 483:
```suggestion
        "Restoring the snapshot. Depending on the amount of data you have, this might take a very long time."
```

croud/clusters/commands.py
```diff
@@ -442,6 +442,67 @@ def clusters_set_backup_schedule(args: Namespace) -> None:
...
+        body["source_cluster_id"] = args.source_cluster_id
+    if args.tables:
+        tables = args.tables.strip().split(",")
+        body["tables"] = [x.strip() for x in tables]
+
+    client = Client.from_args(args)
+    data, errors = client.post(url, body=body)
+
+    if errors:
+        return
```
tomach, line 481: I think like this we do not print any error messages returned by the API? usually there is a `print_response(..., errors=errors, ...)` above or we could just print the error here..

croud/clusters/commands.py
```diff
@@ -442,6 +442,67 @@ def clusters_set_backup_schedule(args: Namespace) -> None:
...
+    )
+
+    print_response(
+        data=data,
+        errors=errors,
+        keys=[
+            "created",
+            "repository",
+            "snapshot",
+            "tables",
```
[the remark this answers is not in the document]
plaharanne: the api returns the full snapshot object. Maybe we should change it there, because it is not really useful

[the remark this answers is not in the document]
juanpardo: Don't we normally return the async operation in non immediate ops? @tomach @plaharanne 
While implementing this I noticed this looks slightly different from other endpoints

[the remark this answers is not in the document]
tomach: Usually we re-fetch the cluster details after the async operation has finished to show the updated data. In this case this doesn't really make sense I think. Maybe it would be enough to have the `Operation completed.` text from `_wait_for_completed_operation()` as the last output?

croud/__main__.py
```diff
@@ -582,6 +584,54 @@
...
+                            Argument(
+                                "--repository", type=str, required=True,
+                                help="The repository that contains the snapshot to be "
+                                     "restored.",
+                            ),
+                            Argument(
+                                "--source-cluster-id", type=str, required=False,
+                                help="The CrateDB cluster ID of the snapshot to be "
+                                     "used belongs to. Must belong to the same "
+                                     "organization than the target cluster."
```
tomach, line 621:
```suggestion
                                     "organization as the target cluster."
```

tomach (approved):
LGTM :+1:

PR #456, opened 2023-08-04
add command to delete users

plaharanne:
## Checklist

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -106,7 +106,7 @@ from croud.subscriptions.commands import (
     subscriptions_list,
 )
 from croud.tools.spinner import HALO
-from croud.users.commands import users_list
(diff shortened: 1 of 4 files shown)


croud/__main__.py
```diff
@@ -1090,6 +1090,17 @@
                     },
                 },
             },
+            "delete": {
+                "help": "Delete the specified user.",
```
SStorm, line 1094: Do we somehow mark these that they are for superusers only?

SStorm (approved):

croud/__main__.py
plaharanne: without `--sudo` it says `==> Error: Permission denied for user.`. I'll add something more specific in the help

PR #467, opened 2023-10-16
add command to show import job progress

plaharanne:
## Checklist

 - [x] Link to issue this PR refers to (if applicable): https://github.com/crate/cloud/issues/1489


docs/commands/clusters.rst
```diff
@@ -726,6 +726,31 @@ Example
...
+
+.. code-block:: console
+
+   sh$ ❯ croud clusters import-jobs progress \
+         --cluster-id e1e38d92-a650-48f1-8a70-8133f2d5c400 \
+         --import-job-id 00de6048-3af6-41da-bfaa-661199d1c106
+   +-----------+-----------+------------------+-----------------+---------------+-----------------------------------------------------------------------------------+
+   |   percent |   records |   failed_records |   total_records |   total_files | files                                                                             |
+   |-----------+-----------+------------------+-----------------+---------------+-----------------------------------------------------------------------------------|
+   |       100 |       891 |                0 |             891 |             2 | {'failed_records': 0, 'name': 'file1.csv', 'records': 800, 'total_records': 891}, |
```
SStorm, line 749: Do you think it would be nicer to print each file as a separate line instead? Maybe show a summary table at the top, and then individual files in a table after that? WDYT?

SStorm (approved):

docs/commands/clusters.rst
plaharanne: I don't like that much to have 2 tables, however I agree, a table for the files only would be nice. I'll add a param `--summary` to show the global values and without it it would display the individual files.
Thanks for the feedback.

plaharanne: or even 2 commands:
- `clusters import-jobs progress files`
- `clusters import-jobs progress`

SStorm: Maybe too much hassle for two commands, I think --summary would be enough ;)

PR #483, opened 2024-01-16
Added ability to schedule sql jobs

Taliik:
## Checklist

 - [ ] Link to issue this PR refers to (if applicable): https://github.com/crate/cloud/issues/1616

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes for croud
 Unreleased
 ==========
 
+- Added support for scheduling sql jobs with the ``scheduled-jobs`` commands.
+
 1.10.1 - 2024/01/11
 ===================
 
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -109,6 +109,12 @@ from croud.projects.users.commands import (
     project_users_remove,
 )
 from croud.regions.commands import regions_create, regions_delete, regions_list
+from croud.scheduledjobs.commands import (
+    create_scheduled_job,
+    delete_scheduled_job,
+    get_scheduled_job_log,
+    get_scheduled_jobs,
+)
 from croud.subscriptions.commands import (
     subscription_delete,
     subscriptions_create,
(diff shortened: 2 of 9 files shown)


croud/gcjobs/commands.py
```diff
@@ -0,0 +1,108 @@
...
+
+    if errors or not data:
+        return
+
+
+@grand_central_jwt_token
+def get_scheduled_job_log(args: Namespace) -> None:
+    client = _get_gc_client(args)
+
+    data, errors = client.get(f"/api/scheduled-jobs/{args.job_id}/log")
```
plaharanne, line 51: No slash at the end?

Taliik: https://github.com/crate/grand-central/blob/365cef5d67d3fea3fa8ebfcf68e57c82dd385d15/api/cratedb/grandcentral/api/scheduled_jobs/views.py#L35

docs/commands/gcjobs.rst
```diff
@@ -0,0 +1,108 @@
+======
+``gc``
+======
+
+The ``gc`` command lets you manage scheduled sql jobs for your cluster.
```
plaharanne, line 5: I'm wondering if the keyword here should be something like`scheduled-sql` instead of `gc`. I guess we could have other GC features in croud later, it might be confusing. idk.

my_table.csv
plaharanne, line 1: Is it used in the tests? I can't find it.

Taliik: thanks for finding this. It appears that when you run the whole test suit, the test `test_export_job_create` writes the `my_table.csv` file but does not clean it up ...

docs/commands/gcjobs.rst
SStorm: I think I agree here, gc is not really a public facing thing...

Maybe `croud scheduled-jobs`?

plaharanne (approved):

PR #492, opened 2024-02-28
Add commands to manage cloud-configurations

tomach:
Add superuser commands to manage cloud configurations
- `croud cloud-configurations list --sudo [--org-id ORG_ID]` lists all available configuration keys/values. Optionally with org specific overrides.
- `croud cloud-configurations get --key KEY --sudo [--org-id ORG_ID]` get the value of a single configuration key. Optionally with org specific overrides.
- `croud cloud-configurations set --key KEY --value VALUE --sudo [--org-id ORG_ID]` set the value of a configuration key globally or create an org override if an `org-id` is set.

## Checklist

 - [x] Link to issue this PR refers to (if applicable): https://github.com/crate/cloud/issues/1682

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -31,6 +31,11 @@ from croud.apikeys.commands import (
     api_keys_edit,
     api_keys_list,
 )
+from croud.cloud_configurations.commands import (
+    cloud_configurations_get,
+    cloud_configurations_list,
+    cloud_configurations_set,
+)
 from croud.clusters.commands import (
     clusters_delete,
     clusters_deploy,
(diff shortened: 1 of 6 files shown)


croud/__main__.py
```diff
@@ -1632,6 +1637,64 @@
             },
         },
     },
+    "cloud-configurations": {
+        "help": "Manage configurations of CrateDB Cloud.",
+        "commands": {
+            "set": {
+                "help": (
+                    "Set a configuration key/value. Command "
```
juanpardo, line 1645: I think it would sound better to write ``This command is for superusers only.``

juanpardo (approved):

PR #505, opened 2024-04-09
upgrade crate docs version

plaharanne:
## Checklist

 - [ ] Link to issue this PR refers to (if applicable):  https://github.com/crate/cloud/issues/1791

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -19,11 +19,11 @@ jobs:
         os: [ubuntu-latest, macos-latest]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.11
       - name: Build docs
         run: |
           cd docs && make check
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,6 +7,8 @@ Unreleased
 
 - Update doc for ``clusters restart-node`` because it is not superuser only anymore.
 
+- Upgrade docs build action Python version to 3.11.
+
 1.11.0 - 2024/03/12
 ===================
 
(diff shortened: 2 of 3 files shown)


docs/requirements.txt
```diff
@@ -1,3 +1,3 @@
 .
-crate-docs-theme
+crate-docs-theme==0.31.0
```
amotl, line 2: Can you remove that version pinning again? Pinning to 0.30.0 might have helped beforehand, but now, most recent 0.31.1 should also improve the situation on older versions of Python. Thanks!

.github/workflows/main.yml
```diff
@@ -19,10 +19,10 @@ jobs:
     strategy:
       matrix:
         python-version:
-        - 3.8
+        - 3.11
```
amotl, line 22: Maybe not bump Python version on the main package yet, and only have this boxed into documentation concerns for now?

Otherwise, you would need to also touch other details, probably bumping mypy over here.
-- https://github.com/crate/croud/blob/1.11.0/setup.py#L67

.github/workflows/release.yaml
```diff
@@ -8,11 +8,11 @@ jobs:
...
       - name: Checkout
-        uses: actions/checkout@v2
-      - name: Set up Python 3.8
-        uses: actions/setup-python@v4
+        uses: actions/checkout@v4
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
         with:
-          python-version: 3.8
+          python-version: 3.11
```
amotl, line 15: Could work, but maybe better omit that one as well, only scoping the PR about documentation matters?

.github/workflows/main.yml
plaharanne: I've already undone that change. thanks :)

.github/workflows/release.yaml
plaharanne: I've already omit that one too. We'll consider upgrading everything an other time. You're right, let's focus on the docs here.

PR #506, opened 2024-04-10
Update to mypy 1.10

amotl:
## About

Update to the most recent version of mypy, in order to support the update to Python 3.11 on a subsequent iteration.

### DIFF
--- a/croud/regions/commands.py
+++ b/croud/regions/commands.py
@@ -138,7 +138,7 @@ def regions_delete(args: Namespace) -> None:
         print_response(
             data=data,
             errors={
-                "message": region_errors.get("message"),
+                "message": region_errors.get("message", "unknown error"),
                 "errors": {"related_resources": region_errors.get("related_resources")},
             },
             output_fmt=get_output_format(args),
--- a/croud/server.py
+++ b/croud/server.py
@@ -57,7 +57,7 @@ class SetTokenHandler(BaseHTTPRequestHandler):
             msg = SetTokenHandler.DUPLICATE_TOKEN_MSG
         else:
             token = query["token"][0]
-            self.server.on_token(token)
+            self.server.on_token(token)  # type: ignore[attr-defined]
             code = 200
             msg = SetTokenHandler.SUCCESS_MSG
         self.send_response(code)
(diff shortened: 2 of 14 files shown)


croud/regions/commands.py
```diff
@@ -138,7 +139,7 @@ def regions_delete(args: Namespace) -> None:
         print_response(
             data=data,
             errors={
-                "message": region_errors.get("message"),
+                "message": cast(str, region_errors.get("message")),
```
amotl, line 142: That might be a viable alternative instead of using a `cast`. wdyt?
```suggestion
                "message": region_errors.get("message", "unknown error"),
```

croud/server.py
```diff
@@ -57,7 +57,7 @@ def do_GET(self):
             msg = SetTokenHandler.DUPLICATE_TOKEN_MSG
         else:
             token = query["token"][0]
-            self.server.on_token(token)
+            self.server.on_token(token)  # type: ignore[attr-defined]
```
amotl, line 60: I've marked this attribute to be ignored. Apparently, the `on_token` is a non-standard extension, where the type checker trips on, but in reality, all just works well?

pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[tool.mypy]
+check_untyped_defs = true
+ignore_missing_imports = true
+implicit_optional = true
+install_types = true
+non_interactive = true
```
amotl, line 6: Well, I used the chance to introduce a `pyproject.toml`, which can absorb more project metadata in the future while we go. I hope you agree.

tests/test_printer.py
```diff
@@ -17,6 +17,8 @@
 # with Crate these terms will supersede the license and you may use the
 # software solely pursuant to the terms of the relevant commercial agreement.
 
+# type: ignore
+
```
amotl, line 21: On `test_printer.py`, mypy emits many type errors, so I figured it would be good to ignore them altogether. Let me know if you think different, then it will probably be a bigger story.

croud/regions/commands.py
plaharanne: Yes, it makes sense to have a meaningful message.

pyproject.toml
plaharanne: sounds good. Fabian is working on migrating our projects to Poetry, I guess he would eventually create it anyway.

amotl: Is your decision on Poetry already fixed? I think classic setuptools/pip is mature enough today so you don't exactly need Poetry, mostly. Also, it is aging a bit.

croud/regions/commands.py
amotl: Hi. I changed the code according to the suggestion. The patch is now ready for review, if you don't have any other objections.

amotl: > Bumping the mypy version should be done in `.github/dependabot.yml` too.

Right. Done.

amotl:
Hi. I've updated this PR according to corresponding suggestions from code review. Can you have a final look?

plaharanne (approved):

PR #543, opened 2024-11-06
Dependencies: Modernize to prepare unlocking more recent Python versions

amotl:
## About
A few adjustments to prepare unlocking Python 3.12 and eventually 3.13.
The patch mainly phases out use of `pkg_resources` and `distutils.util`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ setup(
         "bitmath==1.3.3.1",
         "certifi",
         "colorama==0.4.6",
+        "importlib-metadata; python_version < '3.8'",
         "marshmallow==3.22.0",
         "pyyaml==6.0.2",
         "requests==2.32.3",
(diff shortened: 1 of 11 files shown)


setup.py
```diff
@@ -60,8 +61,11 @@
...
     extras_require={
         "testing": [
-            "tox==3.14.2",
-            "pytest-freezegun==0.4.2",
+            "pytest<9",
+            "pytest-cov<7",
+            "pytest-freezer<0.5",
+            "pytest-random-order<2",
+            "tox<4",
         ],
```
amotl, line 69: FYI: `pytest-freezegun` got superseded by `pytest-freezer`.

- https://github.com/ktosiek/pytest-freezegun/issues/44

tox.ini
```diff
@@ -2,6 +2,6 @@
 envlist = py38,py39,py310,py311
 
 [testenv]
-deps = -r{toxinidir}/test-requirements.txt
+deps = -e{toxinidir}[testing]
```
amotl, line 5: I haven't tested it, but it looks reasonable.

setup.py
```diff
@@ -60,8 +61,11 @@
     ],
     extras_require={
         "testing": [
-            "tox==3.14.2",
-            "pytest-freezegun==0.4.2",
+            "pytest<9",
```
plaharanne, line 64: I think in the past we tried to use `==` if possible instead of `<`.
I don't have a strong opinion about that. @Taliik What do you think about it?

Taliik: I'm fine with more lenient version pinning for testing 👍

PR #556, opened 2025-01-22
Drop support for Python 3.8

tomach:
Python 3.8 reached EOL. and it is causing dependency upgrades to fail GH-546, GH-551

## Checklist

 - [ ] Link to issue this PR refers to (if applicable): Fixes #???

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -50,7 +50,6 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
-        - '3.8'
         - '3.9'
         - '3.10'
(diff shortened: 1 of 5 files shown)


plaharanne (approved):
You could also refresh DEVELOP.rst that contains a lot of old examples with python 3.6 and 3.7. Otherwise, it looks good.

tomach:
> You could also refresh DEVELOP.rst that contains a lot of old examples with python 3.6 and 3.7. Otherwise, it looks good.

good catch, thank you! I've updated `DEVELOP.rst`

PR #564, opened 2025-04-26
Fix duplicate JSON output in `croud clusters deploy`

amotl:
## Problem
`croud clusters deploy` yielded duplicate output, which is detrimental when using `--output-fmt=json`, because the output can not be parsed by downstream programs.

## References
- https://github.com/crate/cratedb-toolkit/pull/81#pullrequestreview-1737914416

### DIFF
--- a/croud/clusters/commands.py
+++ b/croud/clusters/commands.py
@@ -114,14 +114,13 @@ def clusters_deploy(args: Namespace) -> None:
         return
 
     data, errors = client.post(f"/api/v2/organizations/{org_id}/clusters/", body=body)
-    print_response(
-        data=data,
-        errors=errors,
-        keys=["id", "name", "fqdn", "url"],
-        output_fmt=get_output_format(args),
-    )
-
     if errors or not data:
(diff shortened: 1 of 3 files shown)


croud/clusters/commands.py
```diff
@@ -114,14 +114,13 @@ def clusters_deploy(args: Namespace) -> None:
...
     data, errors = client.post(f"/api/v2/organizations/{org_id}/clusters/", body=body)
-    print_response(
-        data=data,
-        errors=errors,
-        keys=["id", "name", "fqdn", "url"],
-        output_fmt=get_output_format(args),
-    )
-
     if errors or not data:
+        print_response(
```
tomach, line 118: no strong opinion about it. but if we change it, we also have to update the example output in the [docs](https://github.com/crate/croud/blob/master/docs/commands/clusters.rst)

amotl: Thanks. I've amended the patch accordingly.

tomach (approved):

Issue #257, opened 2019-12-20
Account for edge-case when login-token is an empty string


MarkusH:
After a login, the user is redirected to `http://localhost:8400/?token=...` where we take the token to write it into the config file. We should account for the value of `token` to be empty.

amotl:
This might be a corresponding fix.

- GH-565

PR #565, opened 2025-04-26
Client: Account for edge-case when login token is an empty string

amotl:
## About
What the title says.

## References
- GH-257


tomach (approved):
:+1:

PR #572, opened 2025-05-13
Add default disk size for croud clusters deploy

hlcianfagna:
## Checklist

 - [ ] Link to issue this PR refers to (if applicable): Fixes #???


hlcianfagna:
The default disk size depends on the product https://github.com/crate/cloud-api/blob/master/data/products/clusters.yaml

PR #576, opened 2025-06-09
Dependencies: Relax version pinning

amotl:
## About
Locking down transitive dependencies to patch release versions and applying them as main package runtime dependency constraints isn't always the best option when aiming to cooperate with other packages.

## Problem
This became apparent with dependencies of `marshmallow` and `requests`, making it difficult to use croud in a wider ecosystem together with other packages in the same Python environment, either as a library, or just installed side-by-side, due to package version conflicts. [^1]

> [...] because croud==1.13.0 depends on marshmallow==3.22.0 and ingestr==0.13.52 depends on marshmallow==3.26.1 [...], installing both together doesn't work.

## Thoughts
Application builds of croud might take a different route, e.g. by using dedicated `requirements.txt` files that include the whole shebang, based on using tools like `uv pip compile` to enumerate dependencies thoroughly, including transitive ones.

## References
- https://github.com/bruin-data/ingestr/pull/239


surister (approved):
relaxing cup of dependencias con leche

PR #578, opened 2025-06-09
CI: Stop building `docs` on `macos-latest`

amotl:
## About
It saves a few cycles on the linkchecker, both on GHA and on remote web servers.

## References
- https://github.com/crate/crash/pull/470
- https://github.com/crate/crash/issues/469

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,7 +16,7 @@ on:
 # Cancel in-progress jobs when pushing to the same branch.
 concurrency:
   cancel-in-progress: true
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
 
 jobs:
   documentation:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
 
 concurrency:
   cancel-in-progress: true
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
 
 jobs:
   lint:


amotl:
@surister, @bmunkholm: @coderabbitai is recommending per nitpick comment:

> **Review concurrency grouping**
> Great addition of concurrency cancellation. Consider switching to `${{ github.ref_name }}` instead of `${{ github.ref }}` in the group key to avoid the `refs/heads/` prefix in the concurrency group name.

surister (approved):

bmunkholm:
@amotl Why introduce to run it nightly as well?

amotl:
### Why?
> Right, but does that mean that we need that? [...] What is it that can change outside the repo that makes a build needed? Is it links/references?

Exactly. The purpose is to convey link checker errors I guess. Without submitting PRs, we would otherwise not know about them. This holds true for both kinds docs and code.

### Antipattern?

> What is it that can change outside the repo that makes a build needed?

> Some of our code depends on the outside world (some even on stuff that is still fluctuating).
> -- https://forum.gitlab.com/t/regularly-rerun-ci-tests/33171

> Building daily seems like an anti pattern to me.

A lot of people are writing about it, typical CI software is offering it as a feature, and from my experience it is common practice to catch errors induced by the environment, by off-by-one situations, and other shenanigans.

- [Continuous Integration](https://martinfowler.com/articles/continuousIntegration.html)
- [GitHub](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule), [GitLab](https://docs.gitlab.com/ci/pipelines/schedules/), [Jenkins](https://www.baeldung.com/ops/jenkins-job-schedule), [CircleCI](https://circleci.com/docs/scheduled-pipelines/)

PR #597, opened 2026-03-09
Added support for DynamoDB data jobs

juanpardo:
Added support for DynamoDB data jobs

## Checklist

 - [ ] Link to issue this PR refers to (if applicable): https://github.com/crate/cloud/issues/2795

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -61,6 +61,7 @@ from croud.clusters.commands import (
     get_scheduled_job_log,
     get_scheduled_jobs,
     import_jobs_create_from_azure_blob_storage,
+    import_jobs_create_from_dynamodb,
     import_jobs_create_from_file,
     import_jobs_create_from_s3,
     import_jobs_create_from_url,
@@ -140,20 +141,6 @@ import_job_create_common_args = [
         help="The cluster the data will be imported into.",
     ),
     Argument(
-        "--file-format",
-        type=str,
-        required=True,
-        choices=["csv", "json", "parquet"],
-        help="The format of the structured data in the file.",
-    ),
-    Argument(
-        "--compression",
-        type=str,
-        required=False,
-        choices=["gzip"],
-        help="The compression method the file uses.",
-    ),
-    Argument(
         "--table",
         type=str,
         required=True,
@@ -167,6 +154,23 @@ import_job_create_common_args = [
         help="Whether the table should be created automatically"
         " if it does not exist. If true new columns will also be added when the data"
         " requires them.",
+    ),
+]
+
+import_job_create_common_file_args = [
+    Argument(
+        "--file-format",
+        type=str,
+        required=True,
+        choices=["csv", "json", "parquet"],
+        help="The format of the structured data in the file.",
+    ),
+    Argument(
+        "--compression",
+        type=str,
+        required=False,
+        choices=["gzip"],
+        help="The compression method the file uses.",
     ),
     Argument(
         "--transformations",
@@ -828,7 +832,8 @@ command_tree = {
                                         help="The URL the import file will be read "
                                              "from."
                                     ),
-                                ] + import_job_create_common_args,
+                                ] + import_job_create_common_args
+                                  + import_job_create_common_file_args,
                                 "resolver": import_jobs_create_from_url,
                             },
                             "from-file": {
@@ -852,7 +857,8 @@ command_tree = {
                                              "Please note the file will become visible "
                                              "under ``croud organizations files list``."
                                     ),
-                                ] + import_job_create_common_args,
+                                ] + import_job_create_common_args
+                                  + import_job_create_common_file_args,
                                 "resolver": import_jobs_create_from_file,
                             },
                             "from-s3": {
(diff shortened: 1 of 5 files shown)


amotl:
I would also like to add data loading support for DynamoDB to the SDK. This patch makes it easy to learn about relevant ingredients. Thank you! I am looking forward to the upcoming release.

- https://github.com/crate/cratedb-toolkit/issues/692

croud/clusters/commands.py
```diff
@@ -200,6 +200,28 @@ def import_jobs_create_from_s3(args: Namespace) -> None:
...
+    extra_body = {
+        "dynamodb": {
+            "region": args.aws_region,
+            "table_name": args.dynamodb_table,
+            "secret_id": args.secret_id,
+        },
+    }
+    if args.endpoint:
+        extra_body["dynamodb"]["endpoint"] = args.endpoint
+    if args.kinesis_stream_name:
```
tomach: should there be a validation that `kinesis_stream_name` is required if ingestion_type is `IMPORT_AND_CDC` or `CDC_ONLY`?

tests/commands/test_clusters.py
```diff
@@ -1601,6 +1601,74 @@ def test_import_job_create_from_s3(mock_request):
...
 
 
+@mock.patch.object(
+    Client,
+    "request",
+    return_value=(
+        {
+            "id": "1",
+            "status": "SUCCEEDED",
+            "progress": {"records": 27, bytes: 4096, "message": "success"},
```
tomach:
```suggestion
            "progress": {"records": 27, "bytes": 4096, "message": "success"},
```

croud/clusters/commands.py
juanpardo: Good point, I just added a check

croud/clusters/commands.py
```diff
@@ -200,6 +200,32 @@ def import_jobs_create_from_s3(args: Namespace) -> None:
...
+            "secret_id": args.secret_id,
+        },
+    }
+    if args.endpoint:
+        extra_body["dynamodb"]["endpoint"] = args.endpoint
+    if args.ingestion_type:
+        extra_body["ingestion_type"] = args.ingestion_type
+    if "CDC" in extra_body.get("ingestion_type", ""):
+        if not args.kinesis_stream_name:
+            raise Exception(
```
tomach: Raising a plain `Exception` here will print a full stack trace I think? maybe raise a `ValueError` so the error looks a bit better? sth like
```
raise ValueError(
    "--kinesis-stream-name must be set when using CDC ingestion types."
)
```

docs/commands/data-import-export.rst
```diff
@@ -172,6 +172,29 @@ Example
...
+``clusters import-jobs create from-dynamodb``
+=============================================
+
+.. argparse::
+   :module: croud.__main__
+   :func: get_parser
+   :prog: croud
+   :path: clusters import-jobs create from-dynamodb
+
+.. code-block:: console
```
plaharanne, line 191: You need to change the titles level.
I've also added a note about the waiting to finish like the others, but actually I don't know how it behave. I let you fix the text I suggested if it is not accurate.

```suggestion
``clusters import-jobs create from-dynamodb``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

.. note::

   For IMPORT_ONLY, this command will wait for the operation to finish or fail.

.. argparse::
   :module: croud.__main__
   :func: get_parser
   :prog: croud
   :path: clusters import-jobs create from-dynamodb
  
Example
^^^^^^^

.. code-block:: console
```

croud/clusters/commands.py
juanpardo: Raising any exception leads to a stack trace, so I had to end up using a controlled error exit

docs/commands/data-import-export.rst
juanpardo: Fixed this, thanks!

docs/commands/data-import-export.rst
```diff
@@ -172,6 +172,36 @@ Example
...
 
+``clusters import-jobs create from-dynamodb``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+.. note::
+
+   For IMPORT_ONLY, this command will wait for the operation to finish or fail.
+   When --ingestion-type is set to CDC_ONLY or IMPORT_AND_CDC, the command will not finish
+   and even when the last CDC event is processed, it will remain waiting for new CDC events to come.
```
plaharanne, line 183: I'm wondering if we should change a bit the command behavior when it's CDC (and for MongoDB as well) to make it just start the importJob and not wait at all.

juanpardo: Would you wait for the import part to finish in an `IMPORT_AND_CDC` job though? Happy to discuss what the best behaviour would be for MongoDB and DynamoDB, but perhaps we can do it in a separate issue :)

plaharanne: You're right, it can be done in a separate issue. Will you create it?

juanpardo: Done!
https://github.com/crate/cloud/issues/2878

croud/__main__.py
```diff
@@ -168,6 +155,23 @@
...
+        type=str,
+        required=True,
+        choices=["csv", "json", "parquet"],
+        help="The format of the structured data in the file.",
+    ),
+    Argument(
+        "--compression",
+        type=str,
+        required=False,
+        choices=["gzip"],
```
plaharanne: You added none to the exports but not here
```suggestion
        choices=["gzip", "none"],
```

docs/commands/data-import-export.rst
```diff
@@ -172,6 +172,36 @@ Example
    ==> Info: Done importing 14.73K records
    ==> Success: Operation completed.
 
+``clusters import-jobs create from-dynamodb``
```
plaharanne, line 175: Do you mind also adding the block of documentation for mongodb? In my PR I added the missing command but it looks like I forgot to add it in this file too.

plaharanne: I can add it in a separated PR if you think it doesn't fit here

juanpardo: Sorry but I can't see the MongoDB command for importing. I can only see MongoDB compatibility in the secret creation command but not on the import creation :s

plaharanne: You're right, it's missing. I thought I added it, I might have mixed stuff or forget to merge some files.

juanpardo: No problem ^^'

plaharanne (approved):
👍

PR #598, opened 2026-04-01
Fixed the ingestion type parameter for DynamoDB data jobs

juanpardo:

### DIFF
--- a/croud/__main__.py
+++ b/croud/__main__.py
@@ -908,16 +908,16 @@ command_tree = {
                                     Argument(
                                         "--ingestion-type", type=str,
                                         choices=[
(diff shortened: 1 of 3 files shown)


plaharanne (approved):
Can you add it to the changelog too?

juanpardo:
> Can you add it to the changelog too?

The changelog already has a `Added support for DynamoDB data jobs` entry, and hasn't been released yet, that's why I thought it would feel redundant to have a feature and a fix for the feature in the same release changelog entry. What do you think @plaharanne ?

plaharanne:
Hey @juanpardo I think you can merge this pull request. It is opened for a while

